PR #30, opened 2019-05-09
handling missing values

sipah00:


src/interpolation_utils.jl
```diff
@@ -54,3 +54,19 @@ function OneCompartmentPK_cmax(A)
   t = OneCompartmentPK_tmax(A)
   p[1] * p[2] * p[3] * (exp(-p[4]*t) - exp(-p[1]*t)) / (p[5] * (p[1] - p[4]))
 end
+
+# helper function for data manipulation
+function munge_data(u::AbstractVector, t)
+  df = DataFrame(u = u, t = t)
```
ChrisRackauckas: why does this need to become a dataframe?

sipah00: we can do this by writing loops (iterate over `t` and check if is missing then drop `t` and corresponding `u` and again similarly  we have to do it iterating over `u` and ....) so I thought this would be an inefficient approach and after making `dataframe`  we don't need loops and can just call `dropmissing` and also it would be efficient if dropmissing on DataFrame is using parallel computing

sipah00: may be I have missed a simpler and efficient approach

ChrisRackauckas: DataFrames.jl is generally slow, but this is probably good enough.

PR #46, opened 2019-09-06
Fix dir=:left in ZeroSpline

andreasnoack:
and fix `dir=:left` to include the data points in the interpolation. I've also added some comments describing `dir`. Adding proper documentation will require more effort. Just wanted this in before Sunday.

### DIFF
--- a/src/caches/interpolation_caches.jl
+++ b/src/caches/interpolation_caches.jl
@@ -45,7 +45,7 @@ end
 struct ZeroSpline{uType,tType,dirType,FT,T} <: AbstractInterpolation{FT,T}
   u::uType
   t::tType
-  dir::Symbol
+  dir::Symbol # indicates if value to the $dir should be used for the interpolation
   ZeroSpline{FT}(u,t,dir) where FT = new{typeof(u),typeof(t),typeof(dir),FT,eltype(u)}(u,t,dir)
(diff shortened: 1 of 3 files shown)


src/caches/interpolation_caches.jl
```diff
@@ -45,11 +45,11 @@ end
 struct ZeroSpline{uType,tType,dirType,FT,T} <: AbstractInterpolation{FT,T}
   u::uType
   t::tType
-  dir::Symbol
+  dir::Symbol # indicates if value to the $dir should be used for the interpolation
   ZeroSpline{FT}(u,t,dir) where FT = new{typeof(u),typeof(t),typeof(dir),FT,eltype(u)}(u,t,dir)
 end
 
-function ZeroSpline(u,t;dir=:left)
+function ZeroSpline(u,t;dir=:right)
```
ChrisRackauckas: should we do this in this library? I am unsure as to whether this is what people would expect.

andreasnoack: For now, I've kept the `dir=:left` default so this PR just fixes the issue with inclusion of the data points when `dir=:left`.

Maybe we should force the user to make a decision about which point to use and simply remove the default. That's for a separate PR, though.

PR #49, opened 2019-10-02
Add Akima interpolation method

helgee:
First of all, thanks for the package! This is exactly what I needed!

This is a quick and dirty port of my Akima interpolation implementation from [EarthOrientation.jl](https://github.com/JuliaAstro/EarthOrientation.jl/blob/master/src/akima.jl). Let me know if you want me to change anything!


ChrisRackauckas:
Awesome! Thanks!

PR #51, opened 2019-11-12
Make BSplines work with integer support points

helgee:
I was just tripped up by the fact that you cannot use integer vectors as support points for the BSplines since you are getting `InexactError`s.

This fixes it by making the temporaries float vectors. Don't know whether this is sensible 🤷‍♂

### DIFF
--- a/src/caches/interpolation_caches.jl
+++ b/src/caches/interpolation_caches.jl
@@ -159,8 +159,8 @@ function BSplineInterpolation(u,t,d,pVecType,knotVecType)
   u, t = munge_data(u, t)
   n = length(t)
   s = zero(eltype(u))
-  p = zero(t)
-  k = zeros(eltype(t),n+d+1)
+  p = float(zero(t))
+  k = float(zeros(eltype(t),n+d+1))
   l = zeros(eltype(u),n-1)
   p[1] = zero(eltype(t))
   p[end] = one(eltype(t))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
I don't like silently changing people's types because you never know what they want. For example, this would make using rationals not work.

helgee:
I agree in principle but this is just changing the type of the temporaries and the algorithm seems to expect them to be floats. Also my branch works with rationals, i.e. returns the same float result if both vectors are rationals 😉

In the end, I am not sure whether this is trading off genericity.

ChrisRackauckas:
>returns the same float result if both vectors are rationals

But it doesn't return the rational result, and not all rationals have a float result.

helgee:
Yes, I know but so does `master`.
```julia
julia> t = [1//8,1//3,1//2]
3-element Array{Rational{Int64},1}:
 1//8
 1//3
 1//2

julia> A = BSplineInterpolation(t,t,2,:ArcLen,:Average)
6-element BSplineInterpolation{Array{Rational{Int64},1},Array{Rational{Int64},1},Array{Rational{Int64},1},Array{Rational{Int64},1},Array{Rational{Int64},1},true,Rational{Int64}}:
 1//8
 1//3
 1//2
 1//8
 1//3
 1//2

julia> A(1//4)
0.25
```
But nevermind, with my patch the code is still not generic and I can live with manually converting the input to float.

ChrisRackauckas:
We should probably fix that on master :)

helgee:
Yeah, sorry. I had only time for this drive-by and no idea how to fix it in a more general way.

PR #52, opened 2019-11-14
add curidx to linearinterpolation and zerospline

ChrisRackauckas:
Avoids extra searching by keeping a current index and checking the current value before doing a full search. Maybe it's best to not do the neighbors either, and just do the current, but I went ahead and did the neighbors for now. Everything is `ifelse` since the computations are cheap, I assume that might be faster but haven't fully tested the speed yet.


ChrisRackauckas:
@YingboMa might have ideas here as well.

ChrisRackauckas:
Improved and finished in https://github.com/PumasAI/DataInterpolations.jl/pull/147

PR #65, opened 2020-06-23
Remove the GaussianProcesses based and Loess interpolations

andreasnoack:
The GaussianProcesses based interpolation has made it difficult to keep this package up to date because it's a heavy dependency that is rarely updated with new compat bounds.

The Loess interpolation has never been properly tested and it seems to produce wrong results in some cases. Eventually, we should add a version based on Loess.jl but for now it's better remove it.


lamorton:
The [tutorial](https://htmlpreview.github.io/?https://github.com/PumasAI/DataInterpolations.jl/blob/v2.0.0/example/DataInterpolations.html) linked from the README still refers to these two methods.

PR #66, opened 2020-06-26
Bump Optim to 0.22

andreasnoack:


ChrisRackauckas:
This should get compathelper

andreasnoack:
Compat helper is enabled but it doesn't run as frequent anymore as it used to and it doesn't look like there is a way to manually trigger GitHub Actions

ChrisRackauckas:
I see. I saw this and assumed the library must've not had it installed.

PR #68, opened 2020-07-06
non-mutating munge_data

d-netto:


andreasnoack:
We used to have a version like that but the result can't be inferred so it's too costly.

ChrisRackauckas:
It needs to be done since Zygote cannot infer through the constructor with this mutation. Maybe the easiest fix for now would be to create new dispatches so that Missing does what it currently does and a non-missing dispatch that just does the no-mutating thing but fast. That will make everything with Zygote work well while not impacting Pumas speeds.

andreasnoack:
It’s fine to rewrite these methods such that they work with Zygote. We just shouldn’t use DataFrames for it because of the inference issue.

d-netto:
I believe the latest commit should be a possible solution without using dataframes

d-netto:
Ok, it's passing.

PR #72, opened 2020-08-05
Add derivatives for the splines

avik-pal:
# TODOs:
- [x] Implement derivatives for all the interpolation methods
- [x] Integrate with ChainRulesCore
- [x] Verify correctness using FiniteDifferences

Follow up from https://github.com/SciML/DiffEqFlux.jl/pull/380

### DIFF
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -100,7 +100,8 @@ function derivative(A::AkimaInterpolation{<:AbstractVector{<:Number}}, t::Number
     i = searchsortedlast(A.t, t)
     i == 0 && return zero(A.u[1])
     i == length(A.t) && return zero(A.u[end])
-    3 * A.u[i] + 2 * A.b[i] + A.c[i]
+    wj = t - A.t[i]
+    @evalpoly wj A.b[i] 2A.c[i] 3A.d[i]
 end
 
 function derivative(A::ConstantInterpolation{<:AbstractVector}, t::Number)
@@ -115,9 +116,8 @@ end
 function derivative(A::QuadraticSpline{<:AbstractVector{<:Number}}, t::Number)
     i = findfirst(x -> x >= t, A.t)
     i == 1 ? i += 1 : nothing
-    Cᵢ = A.u[i-1]
-    σ = 1//2 * (A.z[i] - A.z[i-1]) / (A.t[i] - A.t[i-1])
-    A.z[i-1] + σ
+    σ = 1//2 * (A.z[i] - A.z[i - 1]) / (A.t[i] - A.t[i - 1])
+    A.z[i-1] + 2σ * (t - A.t[i-1])
 end
 
 # CubicSpline Interpolation
@@ -153,10 +153,10 @@ function derivative(A::BSplineApprox{<:AbstractVector{<:Number}}, t::Number)
     idx == 0 ? idx += 1 : nothing
     t = A.p[idx] + (t - A.t[idx]) / (A.t[idx + 1] - A.t[idx]) * (A.p[idx + 1] - A.p[idx])
     n = length(A.t)
-    N = spline_coefficients(A.n, A.d - 1, A.k, t)
+    N = spline_coefficients(A.h, A.d - 1, A.k, t)
     ducum = zero(eltype(A.u))
     for i = 1:(A.h - 1)
-        ducum += N[i + 1] * A.d * (A.c[i + 2] - A.c[i + 1]) / (A.k[i + A.d + 1] - A.k[i + 2])
+        ducum += N[i + 1] * A.d * (A.c[i + 1] - A.c[i]) / (A.k[i + A.d + 1] - A.k[i + 1])
     end
     ducum
 end
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -140,33 +140,3 @@ function derivative(A::CubicSpline{<:AbstractVector{<:Number}}, t::Number)
   dD = -(A.u[i] / A.h[i + 1] - A.z[i] * A.h[i + 1] / 6)
   dI + dC + dD
 end
-
-# BSpline Curve Interpolation
-function derivative(A::BSplineInterpolation{<:AbstractVector{<:Number}}, t::Number)
-  # change t into param [0 1]
-  idx = searchsortedlast(A.t,t)
-  idx == 0 ? idx += 1 : nothing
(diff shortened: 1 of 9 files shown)


avik-pal:
I am finding it hard to get the BSpline derivative right. From my understanding the derivative of a BSpline is another BSpline of lower order. [Reference](https://demonstrations.wolfram.com/TheDerivativeVectorsOfABSpline/)
```julia
N = spline_coefficients(n, A.d - 1, A.k[2:(end - 1)], t)
ducum = zero(eltype(A.u))
for i = 1:(n - 1)
      ducum += N[i] * (A.c[i + 1] - A.c[i]) / (A.k[i + A.d + 1] - A.k[i + 1])
end
ducum * A.d
```
But the gradients seem to be way off.

ChrisRackauckas:
That looks like it might introduce some numerical errors? I think @YingboMa might have comments on this.

avik-pal:
@YingboMa any thoughts on this?

ChrisRackauckas:
Is there a version where BSplines are left off that we can open an issue on to complete later?

YingboMa:
Sorry, I missed this PR. I will review it today.

src/derivatives.jl
```diff
@@ -0,0 +1,162 @@
+function derivative(A::LinearInterpolation{<:AbstractVector{<:Number}}, t::Number)
+    idx = findfirst(x -> x >= t, A.t) - 1
```
YingboMa, line 2: If we assume monotonicity, then this can be
```suggestion
    idx = searchsortedfirst(A.t, t) - 1
```

src/derivatives.jl
```diff
@@ -0,0 +1,162 @@
...
+    idx == 0 ? idx += 1 : nothing
```
YingboMa, line 3: This looks pretty strange.
```suggestion
    idx == 0 && ( idx += 1 )
```

src/derivatives.jl
```diff
@@ -0,0 +1,162 @@
...
+    if idx == length(A.t) - 1
+        i₀ = idx - 1; i₁ = idx; i₂ = i₁ + 1;
+    else
+        i₀ = idx; i₁ = i₀ + 1; i₂ = i₁ + 1;
+    end
+    dl₀ = (2t - A.t[i₁] - A.t[i₂]) / ((A.t[i₀] - A.t[i₁]) * (A.t[i₀] - A.t[i₂]))
+    dl₁ = (2t - A.t[i₀] - A.t[i₂]) / ((A.t[i₁] - A.t[i₀]) * (A.t[i₁] - A.t[i₂]))
+    dl₂ = (2t - A.t[i₀] - A.t[i₁]) / ((A.t[i₂] - A.t[i₀]) * (A.t[i₂] - A.t[i₁]))
+    A.u[:, i₀] * dl₀ + A.u[:, i₁] * dl₁ + A.u[:, i₂] * dl₂
```
YingboMa, line 40:
```suggestion
    @views @. A.u[:, i₀] * dl₀ + A.u[:, i₁] * dl₁ + A.u[:, i₂] * dl₂
```

src/derivatives.jl
```diff
@@ -0,0 +1,162 @@
...
+    tmp = G
+    for i = 1:length(idxs)
+        mult = one(A.t[1])
+        for j = 1:(i - 1)
+            mult *= (A.t[idxs[i]] - A.t[idxs[j]])
+        end
+        for j = (i+1):length(idxs)
+            mult *= (A.t[idxs[i]] - A.t[idxs[j]])
+        end
+        wi = inv(mult)
```
YingboMa, line 59: We probably want to cache the barycentric weights in the `LagrangeInterpolation` structure. Computing the weights takes O(n^2) time and is independent of evaluation. Evaluation takes only O(n) time. Not caching weights renders barycentric Lagrange interpolation as slow as the classical one.

src/derivatives.jl
```diff
@@ -0,0 +1,162 @@
...
+function derivative(A::AkimaInterpolation{<:AbstractVector{<:Number}}, t::Number)
+    i = searchsortedlast(A.t, t)
+    i == 0 && return zero(A.u[1])
+    i == length(A.t) && return zero(A.u[end])
+    wj = t - A.t[i]
+    @evalpoly wj A.b[i] 2A.c[i] 3A.d[i]
+end
+
+function derivative(A::ConstantInterpolation{<:AbstractVector}, t::Number)
+    return isnothing(findfirst(A.t .== t)) ? zero(A.u[1]) : eltype(A.u)(NaN)
```
YingboMa, line 108: You can use `isempty(searchsorted(A.t, t))`

YingboMa:
I can help with the BSpline derivative later today.

YingboMa:
For the BSpline derivative, you should check pg. 72 of the NURBS book https://www.springer.com/gp/book/9783642973857. I am not familiar with BSplines, but I think a bulletproof way of differentiating BSplines is just to differentiate each basis function individually.

avik-pal:
The current implementation was differentiating each basis function itself. I removed the BSpline implementation for the time being so that we can move forward with the PR, and can tackle that later.

YingboMa:
I think you have an off-by-one error in your code, also the derivative of the coordinate transformation is missing.

This gives the correct derivative
```julia
using DataInterpolations, Calculus
ts = 0:4
data = sin.(ts)
A = BSplineInterpolation(data,float.(ts),3,:ArcLen,:Average)
n = length(A.t)
t = 0.1
idx = searchsortedlast(A.t,t)
idx == length(A.t) ? idx -= 1 : nothing
t_ = A.p[idx] + (t - A.t[idx])/(A.t[idx+1] - A.t[idx]) * (A.p[idx+1] - A.p[idx])

Nm1 = DataInterpolations.spline_coefficients(n, A.d-1, A.k, t_)
scale = (A.t[idx+1] - A.t[idx]) * (A.p[idx+1] - A.p[idx])
ducum = zero(eltype(A.u))
for i = 1:(n - 1)
    ducum += Nm1[i+1] * (A.c[i + 1] - A.c[i]) / (A.k[i + A.d + 1] - A.k[i + 1])
end
ducum * A.d * scale
Calculus.derivative(A, t)
```

```julia
julia> ducum * A.d * scale
1.0854013436178458

julia> Calculus.derivative(A, t)
1.0854013436149472
```

ChrisRackauckas:
I rebased this branch back to current master since it was falling behind a bit.

PR #73, opened 2020-08-06
change in diff

d-netto:
Solves issue with SplineLayer in DiffEqFlux.

### DIFF
--- a/src/interpolation_caches.jl
+++ b/src/interpolation_caches.jl
@@ -128,7 +128,7 @@ end
 function CubicSpline(u,t)
   u, t = munge_data(u, t)
   n = length(t) - 1
-  h = vcat(0, diff(t), 0)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
How does `diff` lead to an issue in DiffEqFlux? Is this something that should get an MWE and fixed in Zygote?

d-netto:
> How does `diff` lead to an issue in DiffEqFlux? Is this something that should get an MWE and fixed in Zygote?
`diff` is mutating and throws an error in DiffEqFlux. If `diff` was originally implemented in Zygote, then the permanent fix would be to change it there, I believe. The solution works as a partial fix.

d-netto:
diff is mutating and throws an error in DiffEqFlux. If diff was originally implemented in Zygote, then the permanent fix would be to change it there, I believe. The solution works as a partial fix.

PR #76, opened 2020-09-25
Fix BSpline evaluation error on final point

LHTTIPacheco:
This fixes an error of indexing that appearead when evaluating the BSpline on the final point used for interpolation (`A(t[end])` on interpolation_tests.jl). Possibly caused when `findfirst` was substituted by `searchsortedlast`.


ChrisRackauckas:
Could you add a test that catches this? Thanks!

LHTTIPacheco:
Sure!

PR #85, opened 2021-06-02
use new ChainRulesCore correctly

oxinabox:


ChrisRackauckas (approved):

ChrisRackauckas:
Oh god, I need to change all of these?

https://github.com/SciML/DiffEqSensitivity.jl/blob/master/src/concrete_solve.jl#L188

Right now I have both:

https://github.com/SciML/DiffEqBase.jl/blob/master/src/solve.jl#L290-L308

but I'm scared to remove the Zygote.@adjoint because it works. Can I get some help making sure that's all good?

oxinabox:
> Oh god, I need to change all of these?
> 
> https://github.com/SciML/DiffEqSensitivity.jl/blob/master/src/concrete_solve.jl#L188

Yeah, those a kinda wrong for ChainRules since they use `nothing` when they should use `NoTangent` and `Tuple` when they should use `Tangent{Tuple}`.

> but I'm scared to remove the Zygote.@adjoint because it works

Yes, right now having the `@adjoint` will be blocking the the `rrule` from ever being hit, which is probably why those errors never were detected.
(though Zygote might gracefully accept `rrules` that violate the convention, Diffractor and Nabla are less likely to)

> Can I get some help making sure that's all good?

Yes sure, but after JuliaCon stuff is all submitted.

PR #93, opened 2021-11-01
Register Interpolations for Symbolics

ChrisRackauckas:
@yingboma do you know why the printing is weird?

```julia
julia> using Symbolics; u = 2.0collect(1:10); t = 1.0collect(1:10);

julia> A = LinearInterpolation(u,t);

julia> @variables t
1-element Vector{Num}:
 t

julia> A(t)
[2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 20.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0](t)

julia> Symbolics.operation(A(t))
20-element LinearInterpolation{Vector{Float64}, Vector{Float64}, true, Float64}:
  2.0
  4.0
  6.0
  8.0
 10.0
  ⋮
  6.0
  7.0
  8.0
  9.0
 10.0

julia> Symbolics.arguments(A(t))
1-element Vector{Num}:
 t
```

### DIFF
--- a/src/DataInterpolations.jl
+++ b/src/DataInterpolations.jl
@@ -31,6 +31,13 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
 
 (interp::AbstractInterpolation)(t::Number) = _interpolate(interp, t)
 
+import Symbolics
+Base.@pure __parameterless_type(T) = Base.typename(T).wrapper
+Base.nameof(x::AbstractInterpolation) = typeof(x)
+Base.isbinaryoperator(::Type{T}) where T <: AbstractInterpolation = false
+(interp::DataInterpolations.AbstractInterpolation)(t::Symbolics.Num) = Symbolics.SymbolicUtils.term(interp,t)
+
+
 export LinearInterpolation, QuadraticInterpolation, LagrangeInterpolation, AkimaInterpolation,
        ConstantInterpolation, QuadraticSpline, CubicSpline, BSplineInterpolation, BSplineApprox, Curvefit
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3' # Minimum Julia version required
+          - '1.6' # Minimum Julia version required
           - '1'
 #          - 'nightly'
         os:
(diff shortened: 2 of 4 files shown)


src/DataInterpolations.jl
```diff
@@ -31,6 +31,13 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
 
 (interp::AbstractInterpolation)(t::Number) = _interpolate(interp, t)
 
+import Symbolics
+Base.@pure __parameterless_type(T) = Base.typename(T).wrapper
```
andreasnoack, line 35: What is this function used for?

ChrisRackauckas: `Array{Float64} -> Array` for the op name.

andreasnoack: But where is it used?

ChrisRackauckas: good point.

src/DataInterpolations.jl
```diff
@@ -31,6 +31,13 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
...
+Base.nameof(x::AbstractInterpolation) = typeof(x)
```
YingboMa, line 36: `nameof` should return `Symbol`

src/DataInterpolations.jl
```diff
@@ -31,6 +31,13 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
...
+Base.isbinaryoperator(::Type{T}) where T <: AbstractInterpolation = false
```
YingboMa, line 37: This is should be removed.

src/DataInterpolations.jl
```diff
@@ -31,6 +31,13 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
...
+(interp::DataInterpolations.AbstractInterpolation)(t::Symbolics.Num) = Symbolics.SymbolicUtils.term(interp,t)
```
YingboMa, line 38: Needs `unwrap` on `t`.

ChrisRackauckas: Yeah these were all handled in https://github.com/PumasAI/DataInterpolations.jl/pull/94

PR #94, opened 2021-11-02
patch interpolations

ChrisRackauckas:


src/DataInterpolations.jl
```diff
@@ -33,9 +33,9 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
 
 import Symbolics
 Base.@pure __parameterless_type(T) = Base.typename(T).wrapper
-Base.nameof(x::AbstractInterpolation) = typeof(x)
+Base.nameof(x::AbstractInterpolation) = Symbol(typeof(x))
 Base.isbinaryoperator(::Type{T}) where T <: AbstractInterpolation = false
```
YingboMa, line 37: This should be removed.

src/DataInterpolations.jl
```diff
@@ -33,9 +33,9 @@ ChainRulesCore.frule((_, _, Δt), ::typeof(_interpolate), A::AbstractInterpolati
 
 import Symbolics
 Base.@pure __parameterless_type(T) = Base.typename(T).wrapper
```
YingboMa, line 35: This is dead code.

PR #99, opened 2021-12-10
WIP:  Regularization

jjstickel:
This is a preliminary implementation of Regularization Smoothing. The main features are included -- please review. If you think this method is a good fit in this package, I have other features and user options that I have in mind to implement.

One item of special note is that smoothing by regularization is not only a global method in terms of the data points, but it is also global with respect to the t-values used for the smoothing. For the typical interpolation usage that I have seen, the interpolation happens simultaneously with the smoothing step rather than after the fact. Nonetheless, I think it is reasonable to setup spline interpolation on the smoothed points for the purposes of a posteriori interpolation, and this is what I have implemented. The literature I am aware of says that regularization smoothing with a 2nd-order derivative for the regularized term (and for regularily spaced data) corresponds directly to cubic-spline interpolation. However, I am not aware of relationships for other cases (different derivative orders, unequal weighting, and such), if they exist at all.


ChrisRackauckas:
Still WIP?

jjstickel:
> Still WIP?

Well, kind-of. Maybe I'm abusing the "WIP" designation. What I mean here is, the PR is ready for review for content and package fit, but I anticipate further edits and additional content before merging (at least doc-strings and tests). Please do review the method I have created and the example (added to ` example/DataInterpolations.jmd`). Thanks!

src/regularization_smooth.jl
```diff
@@ -0,0 +1,109 @@
...
+function RegularizationSmooth(u, t, d=2; λ=1.0, t̂=nothing, alg=:gcv_svd)
+    # TBD:  make use of `munge_data`
+    u, t = munge_data(u, t)
+    λ = float(λ) # in case an integer is given for λ
+    N = length(t)
+    if isnothing(t̂)
+        t̂ = t
+        # identity matrix for A (RegularizationTools `A` matrix, not `A` object of
+        # DataInterpolations)
+        A = Array{Float64}(LA.I, (N, N))
```
ChrisRackauckas, line 57: this should probably be eltype matching

ChrisRackauckas:
I think this should be set as Requires'd because it pulls in quite a few dependencies. Other than that and generalizing the typing, it looks good.

jjstickel:
> I think this should be set as Requires'd because it pulls in quite a few dependencies. Other than that and generalizing the typing, it looks good.

Thanks for the review. I looked up Requires.jl, but I am not completely clear on the usage. Can you point me to a package that would be a good example? BTW, the only new dependency is RegularizationTools.jl which is really not so large and imports relatively quickly.

src/regularization_smooth.jl
jjstickel: Got it. Also now using eltype in `_mapping_matrix`.

ChrisRackauckas:
> RegularizationTools.jl which is really not so large and imports relatively quickly

Interesting. It's dependency tree looks rather large for what it is, mostly thanks to the optimization libraries.

https://juliahub.com/ui/Packages/RegularizationTools/W7b5l/0.5.1?page=1

But I'd be fine with it since it seems to mostly overlap with standard dependencies.

ChrisRackauckas (approved):

PR #103, opened 2021-12-22
add append! for online interpolation

pepijndevos:
This adds an `append!` implementation for a few simple interpolations so that they can be used in an online context.

It should be possible to implement this for a few more interpolations that are not global, such as Akima.
But I'll just put this out there to make sure I'm going in the right direction before doing tricky things to make online Akima work and then have to do it again based on feedback.

Some things I'm not totally sure about:

- maybe we should implement `push!`
- maybe the type bounds are too tight now
- it seems very eager to call the `AbstractVector` method and complain there is no `resize!` method.
- lagrange has an immutable length field
- which methods are amenable to appending data

### DIFF
--- a/src/DataInterpolations.jl
+++ b/src/DataInterpolations.jl
@@ -22,6 +22,7 @@ include("interpolation_methods.jl")
 include("plot_rec.jl")
 include("derivatives.jl")
 include("integrals.jl")
+include("online.jl")
 
 function ChainRulesCore.rrule(::typeof(_interpolate),
                               A::Union{LagrangeInterpolation,AkimaInterpolation,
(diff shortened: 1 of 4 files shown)


ChrisRackauckas (dismissed):

ChrisRackauckas (approved):

pepijndevos:
Thinking about it a bit, `push!` is probably easier to implement for nontrivial interpolations and more generally useful, but could be less efficient than `append!`.

It seems that in the standard library `push!` is implemented in terms of `append!`, so maybe we should copy that approach? Problem is our functions take two things to add so we can't rely on the underlying implementation, unless we do something weird with tuples. Of course we can also implement append in terms of push but that would be less efficient

ChrisRackauckas:
I think this looks good. I wouldn't worry about it. I didn't know `push!` was implemented in terms of `append!`.

Issue #113, opened 2022-04-04
Improve linear interpolation with NaNs


GlenHertz:
If there are NaN value in the u vector then NaNs bleed into the interpolation values more than they should:

```julia
julia> t = 1:5
1:5

julia> u = [-20, -10, NaN, 10, 20]
5-element Vector{Float64}:
 -20.0
 -10.0
 NaN
  10.0
  20.0

julia> [t u]
5×2 Matrix{Float64}:
 1.0  -20.0
 2.0  -10.0
 3.0  NaN
 4.0   10.0
 5.0   20.0

julia> map(t->t => li(t), 1:0.5:5)
9-element Vector{Pair{Float64, Float64}}:
 1.0 => -20.0
 1.5 => -15.0
 2.0 => NaN
 2.5 => NaN
 3.0 => NaN
 3.5 => NaN
 4.0 => 10.0
 4.5 => 15.0
 5.0 => 20.0
```

Above `li(2.)` should equal `-10.0`.

(The code is here: https://github.com/PumasAI/DataInterpolations.jl/blob/master/src/interpolation_methods.jl#L2-L6)

PR #115, opened 2022-04-07
Fix type stability issue with linear interpolation #113

GlenHertz:
The fast path for linear interpolation when no interpolation is needed (time point is provided in vector) returned a different type which caused issues with types/performance with higher level code.

I ran some benchmarks and see no difference in performance.

### DIFF
--- a/src/interpolation_methods.jl
+++ b/src/interpolation_methods.jl
@@ -1,17 +1,13 @@
 # Linear Interpolation
-function __interpolate(u1, u2, t1, t2, t)
-    θ = (t - t1)/(t2 - t1)
-    (1 - θ)*u1 + θ*u2
-end
 function _interpolate(A::LinearInterpolation{<:AbstractVector}, t::Number)
     idx = max(1, min(searchsortedlast(A.t, t), length(A.t) - 1))
     t1, t2 = A.t[idx], A.t[idx+1]
     u1, u2 = A.u[idx], A.u[idx+1]
-    # For type stability of fast paths:
-    T = Core.Compiler.return_type(__interpolate, Tuple{typeof(u1), typeof(u2), typeof(t1), typeof(t2), typeof(t)})
-    t == t1 && return T(u1) # Return exact value if no interpolation needed (eg when NaN at t2)
-    t == t2 && return T(u2) # ... (eg when NaN at t1)
-    __interpolate(u1, u2, t1, t2, t)
+    θ = (t - t1)/(t2 - t1)
+    val = (1 - θ)*u1 + θ*u2
+    t == t1 && return oftype(val, u1) # Return exact value if no interpolation needed (eg when NaN at t2)
+    t == t2 && return oftype(val, u2) # ... (eg when NaN at t1)
+    val
 end
 
 function _interpolate(A::LinearInterpolation{<:AbstractMatrix}, t::Number)
--- a/src/interpolation_methods.jl
+++ b/src/interpolation_methods.jl
@@ -1,7 +1,8 @@
 # Linear Interpolation
 function _interpolate(A::LinearInterpolation{<:AbstractVector}, t::Number)
     if isnan(t)
-        t1 = t2 = eltype(A.t)(NaN)
+        # For correct derivative with NaN
+        t1 = t2 = one(eltype(A.t))
         u1 = u2 = one(eltype(A.u))
     else
         idx = max(1, min(searchsortedlast(A.t, t), length(A.t) - 1))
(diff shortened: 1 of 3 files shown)


GlenHertz:
I'm not sure if I'm going about this the right way but it works.

src/interpolation_methods.jl
```diff
@@ -1,13 +1,18 @@
...
-    t == t1 && return u1 # Return exact value if no interpolation needed (eg when NaN at t2)
-    t == t2 && return u2 # ... (eg when NaN at t1)
-    θ = (t - t1)/(t2 - t1)
-    return (1 - θ)*u1 + θ*u2
-  end
+    # For type stability of fast paths:
+    T = Core.Compiler.return_type(__interpolate, Tuple{typeof(u1), typeof(u2), typeof(t1), typeof(t2), typeof(t)})
+    t == t1 && return T(u1) # Return exact value if no interpolation needed (eg when NaN at t2)
+    t == t2 && return T(u2) # ... (eg when NaN at t1)
+    __interpolate(u1, u2, t1, t2, t)
```
andreasnoack, line 14: It's best to avoid `Core.Compiler.return_type`. It can cause extreme slowdowns when cases where the inputs aren't inferred. Usually, the way this is handled is to compute the general return value and use it for the special cases, i.e. to compute

```
θ = (t - t1)/(t2 - t1)
r = (1 - θ)*u1 + θ*u2
```

before the branches and then have

```julia
t == t1 && return oftype(r, u1)
```

test/interpolation_tests.jl
```diff
@@ -81,6 +81,42 @@ using StableRNGs
     @test A(3.0) == 2.0
     @test isnan(A(3.5))
     @test isnan(A(4.0))
+
+    # Test type stability
+    u = Float32.(1:5)
+    t = Float32.(1:5)
+    A = LinearInterpolation(u, t)
+    @test Core.Compiler.return_type(A, Tuple{Float32}) == Float32
```
andreasnoack, line 89: something like
```suggestion
    @test @inferred(A(1.0f0)) == A(1.0)
```

GlenHertz:
Note this changes how the derivative works.  With the fast-path change the derivative of the LinearInterpolation at the exact time point is zero (from ForwardDiff) while previously it was the derivative of the current point and the next.

src/interpolation_methods.jl
GlenHertz: Yeah, sounds good.  The fast path ends up being no faster anyways and it rarely occurs with interpolations.

andreasnoack:
> Note this changes how the derivative works.

Good point. Hopefully people don't rely too much on the derivative here. If they did then they should probably consider a different interpolation. However, you could also consider restricting the branches to when `isnan(ret)`. That was your original motivation anyway and will leave the derivative the way it used to be.

GlenHertz:
The derivative at the exact point is but undefined but I think it is better to stay consistent with before and if they depend on the derivative they should probably use a better interpolation scheme.   I've seen other tools use the average slope from either side for this but that seems like a different/hack interpolation method to me.

ChrisRackauckas:
>Good point. Hopefully people don't rely too much on the derivative here.

They do more than they expect. For example, the ODE solver will differentiate at those points if using a stiff ODE solver, and it will step there if it's in a `tstop`. So this is actually pretty important. But, we already have `frule` and `rrule` overloads, it's just for ForwardDiff. This fixes potential ForwardDiff issues too, so it's a pretty good change.

ChrisRackauckas (approved):

ChrisRackauckas:
>I've seen other tools use the average slope from either side for this but that seems like a different/hack interpolation method to me.

The derivative on the right would be preferred due to the ODE solver context.

ChrisRackauckas:
@GlenHertz can you do a patch bump?

GlenHertz:
@ChrisRackauckas I'm not sure what a patch bump is but I added another commit for the derivative fix.  

I'm trying to get my code to work with NaNs and I ran into this odd case:

```
t = 0:3; 
u = [0, -2, -1, -2]; 
A = LinearInterpolation(u, t); 
dA = t -> ForwardDiff.derivative(A, t)
dA(NaN) # returns -1.0
```
It is because if NaN is given then the `idx` will be the second last one and it will load real points from the vectors and then it has a derivative.   I'll work on a patch.

GlenHertz:
The tests fail right now because using a NaN input for `t` with Integer vector doesn't compute....I'll get back to this tomorrow.

ChrisRackauckas:
Test that the derivative at a point gives the derivative to the right?

ChrisRackauckas (approved):

PR #121, opened 2022-07-20
Add backward-looking option for quadratic interpolation

DaniGlez:
The current behaviour of QuadraticInterpolation chooses as the three points for the parabolic fit the two closest points to the interpolation point `t`, as well as the next point to the right (whenever possible, since you can't do this for the last subinterval); this is therefore a "forward-looking" behaviour. Another option is just as sensible, a priori: take the previous point to the left whenever possible, representing a "backward-looking" behavior. This PR adds this alternative behaviour as an option, while leaving the forward-looking one as the default if no mode is specified (thus users should see any numerical change at all, unless they explicitly opt in to the backward-looking mode). The proposed interface is 

```julia
I = QuadraticInterpolation(u,t)
I = QuadraticInterpolation(u,t,:Forward) # same outcome as the previous one
I = QuadraticInterpolation(u,t,:Backward)
```

* Personally, I want this feature to be able replicate the behaviour of another package (which implemented it by hand) with DataInterpolations

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataInterpolations"
 uuid = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
-version = "3.9.2"
+version = "3.10.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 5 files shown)


DaniGlez:
As a side effect, this PR also fixes a slight inaccuracy in the current docs, which describe the quadratic interpolation as a parabola that interpolates the three "nearest" points, which is not the case (If it were true, continuity would be broken when switching "nearest" points outside a node).

ChrisRackauckas (approved):

PR #128, opened 2022-11-18
Revert "Merge pull request #99 from jjstickel/regularization"

andreasnoack:
This reverts commit 2f56d6039a235b1bd2ca74332191c2d51d793442, reversing changes made to 4b6fad18b42651f7b94c1455f83fe2696f4da016.

The PR pulled in a GPL dependency and therefore incompatible with the license of this package. The functionality would have to be rewritten without depending on `RegularizationTools`.

cc @jjstickel


README.md
```diff
@@ -16,7 +16,7 @@ locally, use the following Weave commands:
 
 ```julia
 using Weave, DataInterpolations
-weave(joinpath(dirname(pathof(DataInterpolations)), "../example", "DataInterpolations.jmd"), out_path=:doc)
+weave(joinpath(dirname(pathof(DataInterpolations)), "../Example", "DataInterpolations.jmd"), out_path=:doc)
```
AlCap23, line 19:
```suggestion
weave(joinpath(dirname(pathof(DataInterpolations)), "../example", "DataInterpolations.jmd"), out_path=:doc)
```

Typo?

AlCap23 (changes requested):

Project.toml
```diff
@@ -9,7 +9,6 @@ Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
-RegularizationTools = "29dad682-9a27-4bc3-9c72-016788665182"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 
 [compat]
```
AlCap23, line 14: The combat bound needs to be removed.

AlCap23 (changes requested):

Project.toml
andreasnoack: I've pushed an updated version. The tests had also been added in a separate PR so the is now an extra revert commit.

AlCap23 (approved):

ChrisRackauckas:
Instead of a revert, this should be a subpackage. Can we please do this via a subpackage instead of just throwing it away?

andreasnoack:
The code is still in the git history so it shouldn't be too much work to add as a sub-package once somebody has the time to do so. Fixing the license violation is pretty urgent, though, so we'd have to move forward with that right away.

jonCAstickel:
The primary developer of RegularizationTools might consider changing the license. What are typical license options that would be acceptable?

jjstickel:
The primary developer of RegularizationTools might consider changing the license. What are typical license options that would be acceptable?

ChrisRackauckas:
MIT

andreasnoack:
@jjstickel Is there a conclusion with respect to the licensing issue here?

jjstickel:
> @jjstickel Is there a conclusion with respect to the licensing issue here?

This slipped my mind with the Thanksgiving holiday. I emailed Markus (RegularizationTools developer) just now.

jjstickel:
@andreasnoack Markus agreed to switch the license to MIT, and a PR has been merged with that change. He also tagged a new release (v0.6.1).

andreasnoack:
Great. I've reverted this one in https://github.com/PumasAI/DataInterpolations.jl/pull/130

PR #130, opened 2022-11-30
Revert "Revert "Merge pull request #99 from jjstickel/regularization""

andreasnoack:
Reverts PumasAI/DataInterpolations.jl#128

The dependency has be relicensed under MIT, see https://github.com/PumasAI/DataInterpolations.jl/pull/128#issuecomment-1332453418 so we can reenable the functionality here without issues.


ChrisRackauckas:
Add a version bump

andreasnoack:
I don't think we need to make a release. There hasn't been a release while this was reverted. The relicensed version of RegularizationTools is a minor release so we no changes are required here.

ChrisRackauckas (approved):

Issue #142, opened 2023-03-01
Speed up interpolation of many points by using correlated guesses


moble:
I deal with timeseries a lot, so for me interpolation involves constructing the interpolant once with thousands of data points, then interpolating to thousands of (monotonic) time steps, and repeating that interpolation many times.  In particular, for me, the LU decomposition in constructing a cubic spline takes a small fraction of the total time.  (But interpolation is *by far* the largest bottleneck for my larger code.)

At least on my laptop, the step that takes the longest is the call to `searchsortedlast`.  The problem is that each time this is called, it starts with no information — the required index might be anywhere in the input data.  However, if successive interpolations are for correlated `t` values, it can vastly increase the speed to assume correlated indices.  I guess this improves the cache locality a lot, too.  So much so that, at least for my use cases, the code speeds up by a factor of 3 or 4.

FWIW, here's the implementation I threw together with inspiration from Numerical Recipes's `hunt` function and the built-in `searchsortedlast`.

```julia
@inbounds function searchsortedlastcorrelated(v::AbstractVector, x, guess::T)::keytype(v) where T<:Integer
    u = T(1)
    bottom = firstindex(v)
    top = lastindex(v)
    if guess < bottom || guess > top
        # For a bad guess, just jump to standard bisection
        lo = bottom
        hi = top
    else
        increment = u
        lo = guess
        hi = guess+u
        if x ≥ v[lo]  # Hunt upwards
            lo = hi
            while true
                hi = lo + increment
                if hi ≥ top
                    hi = top
                    break
                elseif x < v[hi]
                    break
                else
                    lo = hi
                    increment += increment
                end
            end
        else  # Hunt downwards
            hi = lo
            while true
                lo -= increment
                if lo ≤ bottom
                    lo = bottom
                    break
                elseif x ≥ v[lo]
                    break
                else
                    hi = lo
                    increment += increment
                end
            end
        end
    end
    # Begin standard bisection, à la `searchsortedlast`
    while lo < hi - u
        m = Base.midpoint(lo, hi)
        if x < v[m]
            hi = m
        else
            lo = m
        end
    end
    return lo
end
```
I just loop through the vector of times to which I have to interpolate, keeping track of the index output by this routine, and feeding it back in with each call to interpolation.  With this change, the search drops to taking a fraction of the time that it takes just to do the arithmetic of actually evaluating the spline.  (My next goal is to improve this to be able to use SIMD, along the lines of what happened in #123.)

[...]

PR #147, opened 2023-03-08
Use correlated searching for repeated interpolations

moble:
As suggested in #142, this PR adds the ability to track the last index found when interpolating, and use it as a guess for the next interpolation.  Essentially, the current use of `searchsortedlast`/`first` starts over, doing a binary seach over the entire `t` array every time the interpolator is called.  If we track the last index found, then the next time the interpolator is called, we can first do an *expanding* binary search to bracket the desired value, then call `searchsortedlast`/`first` as usual.  When successive values are close together, this means that the searches are much faster.

While the cost of constructing the interpolator does not change, tracking the last index results in a best-case speedup of ~2.7x for `CubicSpline`, when successive values are close together (and a little higher for simpler interpolators).  In the worst case (where successive values are always on opposite ends of the vectors), it can result in a ~15% slowdown due to the unhelpful expanding binary search at the beginning.  However, the original approach of not tracking the index at all is also still available; it now involves essentially one extra `if` statement, which seems to be lost in the timing noise.

The interface — both the public and `_interpolate` parts — should be backwards compatible, but have optional extra arguments for the index.  I've also added a method `(interp::AbstractInterpolation)(t::AbstractVector)`, along with an in-place mutating version, which are features that I have felt were sorely lacking.

I extended this tracking to the `derivative` interface, but left the `integral` interface alone because I don't know that it would be much of an improvement, and it starts to get a little complicated.  There were also a few things like `LagrangeInterpolation` that I left alone because they use a different approach.


src/interpolation_utils.jl
```diff
@@ -57,3 +57,68 @@ function munge_data(U::StridedMatrix, t::AbstractVector)
...
+        u += u
+      end
+    end
+  end
+  return lo, hi
+end
+
+for s in [:searchsortedfirst, :searchsortedlast]
+  @eval begin
+    function Base.$s(v::AbstractVector, x, guess)
```
ChrisRackauckas, line 119: Extending `searchsortedfirst` is type piracy. We should just make a new function.

moble: Renamed to `searchsortedfirstcorrelated` (and `...last...`)

ChrisRackauckas:
This looks good to go, just fix the merge conflict (it moved files) and this can merge. It's breaking because it changes the derivative handling, but this will get merged into the v4 release which is building up for next week, so that's totally fine.

moble:
Thanks.  I've rebased and added some description to the README.

For reference, [here's the script](https://gist.github.com/moble/8c3c3369324f0597045db09a6575ad35) I used to benchmark the new approach versus the old.

moble:
Also, I think this can close #52 and #142.

PR #155, opened 2023-03-21
remove some unneded strong dependencies

KristofferC:
These are either loaded in 1.9 using the extension system or on 1.8 using Requires. In no case are they needed to be in [deps]. 

This avoids installing these dependencies (and adding them to sysimages) in Julia 1.8.


ChrisRackauckas (approved):

ven-k:
Noted

PR #161, opened 2023-05-13
fix ForwardDiff compatibillity with respect to coefficients

xzackli:
This PR enables ForwardDiff derivatives of interpolation results with respect to interpolator inputs, i.e.

```julia
using DataInterpolations, ForwardDiff

# elaborate way of writing f(x) = x²
function f(c)
    xs = 0.0:2.0:4.0
    ys = [c * x for x in xs]
    itp = CubicSpline(ys, xs)
    return itp(c)
end

ForwardDiff.derivative(f, 2.0)
```

In the quadratic and cubic spline interpolators, the column vector `d` in the tridiagonal linear solve `tA\d` has element type `Real` when the coefficients are dual types. This happens because the first element of `d` will not be a dual type, due to the edge case.

https://github.com/PumasAI/DataInterpolations.jl/blob/7a4b5f2773b3957ca79adae63bf6dbfd9aee2a2f/src/interpolation_caches.jl#L125

To be as general as possible, this PR determines the type of the *second* element of `d` and then generates a zero element from it. I've included a small regression test.

### DIFF
--- a/src/interpolation_caches.jl
+++ b/src/interpolation_caches.jl
@@ -122,7 +122,11 @@ function QuadraticSpline(u::uType,t) where {uType<:AbstractVector{<:Number}}
   d_tmp = ones(eltype(t),s)
   du = zeros(eltype(t),s-1)
   tA = Tridiagonal(dl,d_tmp,du)
-  d = map(i -> i == 1 ? 0 : 2//1 * (u[i] - u[i-1])/(t[i] - t[i-1]), 1:s)
+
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
I like it, but it looks like tests are failing.

xzackli:
Sorry about that! The test I included was not quite right, and it should be fixed now.

ChrisRackauckas (approved):

ChrisRackauckas (approved):

PR #173, opened 2023-09-13
Fix example

sathvikbhagavan:


sathvikbhagavan:
Also, the docs mentioned in the README should be updated as it points to an older version. 

But I don't see any CI for docs?

ChrisRackauckas:
We should make this into a standard documenter documentation instead.

docs/src/index.md
```diff
@@ -0,0 +1,15 @@
...
+
+-  Linear Interpolation
+-  Quadratic Interpolation
+-  Lagrange Interpolation
+-  Zero Spline
+-  Quadratic Spline
+-  Cubic Spline
+-  BSpline Global Curve Interpolation
+-  BSpline Global Curve Approximation
+-  Curve fitting
```
ChrisRackauckas, line 15: Use the README? There's a lot more information that you can just copy-paste.

sathvikbhagavan: Yes, done in [c75828a](https://github.com/PumasAI/DataInterpolations.jl/pull/173/commits/c75828a02ffe6c1d6be75f900dcb190031bf33d8)

ChrisRackauckas:
While you're at it, can you add a section that documents the interface, i.e. how to interact with an interpolation `i(t)` but also calculate derivatives and integrals? Then I think we're good to go.

ArnoStrouwen:
When this gets merged in. Does this go into the SciML docs -> dev tools -> 3rd party num utils?

ChrisRackauckas:
I would treat it as non-third party. It's really weird it's not in SciML.

sathvikbhagavan:
Any reason we can't move this to SciML?

ChrisRackauckas:
It should.

ChrisRackauckas (approved):

PR #178, opened 2023-09-23
Docs1

ArnoStrouwen:


ChrisRackauckas:
@sathvikbhagavan do you know what introduced that test error?

ArnoStrouwen:
It is not uncommon for the formatter to eat something on the first format. I'll have to go through it in detail.

sathvikbhagavan:
Formatting is not issue here.
QuadGK@2.9 seems to be the problem.  Tests pass with QuadGK@2.8.
Tried this in https://github.com/SciML/DataInterpolations.jl/pull/179

ChrisRackauckas:
Can you upper bound, isolate and open an issue?

stevengj:
Should be fixed shortly.

The source of the problem was a dispatch snafu because your `LinearInterpolation` is a subtype of `AbstractVector`, which seems a bit weird to me.  (The elements of the vector are apparently the abscissae followed by the ordinates?)  This behavior doesn't seem to be documented — are you sure you want to expose it?

PR #193, opened 2023-10-16
Add extrapolation keyword

sathvikbhagavan:


src/derivatives.jl
```diff
@@ -1,4 +1,7 @@
-derivative(A, t) = derivative(A, t, firstindex(A.t) - 1)[1]
+function derivative(A, t)
+    ((t < A.t[1] || t > A.t[end]) && !A.extrapolate) && error("Cannot extrapolate!")
```
ChrisRackauckas, line 2: we should make this an error type and throw it.

sathvikbhagavan: Addressed in [d45c36f](https://github.com/SciML/DataInterpolations.jl/pull/193/commits/d45c36f924c67c9860b4ffea4eb5262df554bd3a)

ChrisRackauckas (approved):

ChrisRackauckas:
It might be nice to turn off extrapolation by default. Worth chatting about.

sathvikbhagavan:
> It might be nice to turn off extrapolation by default

Yeah, it might be better as the expected behaviour of any interpolation should be within the range of time points provided. 
Extrapolations should be allowed only if needed. Should I make a PR for changing that?

ChrisRackauckas:
yes

PR #197, opened 2023-10-27
fix(bounds): allow interpolation on bounds of time

DhairyaLGandhi:
Currently if one tries to call interpolation for the extremities of the timespan for which the splines are fit, we get the `ExtrapolationError` whereas it isn't actually doing extrapolation. This relaxes the checks for where we query to include those bounds.

Perhaps better checks are needed for when we query with something not time (i.e not linear, sorted, or monotonic)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataInterpolations"
 uuid = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
-version = "4.4.0"
+version = "4.4.1"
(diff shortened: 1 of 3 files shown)


sathvikbhagavan:
I am not sure I am following you. It works at the boundary points and this PR would make it error on those points.

Can you provide an MWE where that is not the case?

DhairyaLGandhi:
Yeah, I was mistaken about the order of operation here.

PR #198, opened 2023-10-27
Add a fast path for evenly-spaced abscissae

DaniGlez:
Currently, all interpolation methods but Lagrangian perform a two-step procedure to determine the index of the datapoints / spline weights to use in the corresponding expression: first, a call to `bracketstrictlymontonic` to get a bracket of the viable indices and then a call to `searchsortedfirst` (respectively, `searchsortedlast`) using the bracket as an initial guess¹. `bracketstrictlymontonic` performs an accelerating expansion of the bracket from a size 1 integer interval to the corresponding side; however, this behavior is implicitly disabled right now, because the provided initial guess is `0` which triggers a bailout path in which the function essentially does nothing.

The current PR enables this behavior with a linear guess. This is a "greedy" approach that aims to provide faster performance for cases where the data is equispaced (many cases, in practice) or nearly so, with hopefully a small penalty otherwise. The first goal is clearly met by the PR (puzzingly, it's even faster than the range shortcut added in #192, not sure why really), while the second one is not, at the moment (see benchmarks in the subsequent comment).

¹ As an additional note, looking into `Base` code it seems that the bracket is indeed expected to contain the "correct" solution, which would require a stronger guarantee than the one given in the `bracketstrictlymontonic` docstring (its behavior might provide it, though? not 100% sure).


DaniGlez:
Linearly equispaced data coordinates:

image

image

image

```julia
using BenchmarkTools, DataInterpolations

# Equispaced
begin
    exec_times = map(3:8) do x
        N = 10^x
        u = rand(N)
        tmax = exp(rand() * 10)
        t = (collect(1:N) .- 1) ./ (N - 1) .* tmax
        li = LinearInterpolation(u, t)
        b = @benchmark $li(rand() * $tmax)
        mean(b.times)
    end
    @show exec_times
end

# Log-distributed
begin
    logdist_times = map(3:8) do x
        N = 10^x
        u = rand(N)
        tmax = exp(rand() * 10)
        t = (exp.((collect(1:N) .- 1) ./ (N - 1)) .- 1)
        t .*= (tmax / t[end])
        li = LinearInterpolation(u, t)
        b = @benchmark $li(rand() * $tmax)
        mean(b.times)
    end
    @show logdist_times
end

# Irregularly distributed
begin
    randist_times = map(3:8) do x
        N = 10^x
        u = rand(N)
        tmax = exp(rand() * 10)
        tdiff = rand(N)
        t = cumsum(tdiff .- first(tdiff))
        t .*= (tmax / t[end])
        li = LinearInterpolation(u, t)
        b = @benchmark $li(rand() * $tmax)
        mean(b.times)
    end
    @show randist_times
end

# ranges

begin
    range_times = map(3:8) do x
        N = 10^x
        u = rand(N)
        tmax = exp(rand() * 10)
        t = ((1:N) .- 1) ./ (N - 1) .* tmax
        li = LinearInterpolation(u, t)
        b = @benchmark $li(rand() * $tmax)
        mean(b.times)
    end
    @show range_times
end
```

DaniGlez:
It's clear that the log-distributed case is quite bad for the PR with large amounts of data (or, potentially, just a smaller cache; I'm on a Ryzen 7800 X3D, which is an outlier in that sense), probably because it has to look up the coordinate array during both the bracketing and the `searchsorted`. Thus, a potential "best-of-both-worlds" alternative is to replace the bracketing by a simple check for the happy path, followed by a call to `searchsorted*` using the bisected side as bracket otherwise.

ChrisRackauckas:
> It's clear that the log-distributed case is quite bad for the PR with large amounts of data (or, potentially, just a smaller cache; I'm on a Ryzen 7800 X3D, which is an outlier in that sense), probably because it has to look up the coordinate array during both the bracketing and the searchsorted. Thus, a potential "best-of-both-worlds" alternative is to replace the bracketing by a simple check for the happy path, followed by a call to searchsorted* using the bisected side as bracket otherwise.

that would be good. It would be nice if during construction time there was a type you can set to choose different search behaviors. That would also be a way to balance it.

ChrisRackauckas:
One note is that this PR should get updated to call the functionality which has moved to https://github.com/SciML/FindFirstFunctions.jl. 

As a way to finish it, what about the following. Calculate `x = var(diff(t))` if close to linear `x` is sufficiently small:

```julia
using Statistics
var(diff(1:6)) # 0.0
var(diff(exp.(1:6))) # 10773.646024936865
```

we would just need a tunable cutoff point here then, and we can flip on "fastmode" if the constructed interpolation satisfied the criteria.

DaniGlez:
Sounds good (I was similarly thinking on using a threshold based on testing with randomly-spaced abscissae, though not based on variance). I have some doubts regarding implementation: do you think it would be best to add a field or type parameter to every interpolant cache recording that test, or do you envision some alternative?

ChrisRackauckas:
I think a field with a bool is fine. Runtime checking that is pretty cheap.

DaniGlez:
Revisiting this topic, I added a simple mechanism along the lines of what we discussed. In particular:

- At interpolant creation time, we check for "quasilinearity" by the simple and cheap mechanism of comparing the middle element of the array of abscissae to the average of the first and last element, with a relative threshold of 10%. We store that into a flag.
- At runtime, if the flag is set, we guess the linear index instead of the current mechanism where there's a cached index
- This seems to be Pareto superior on the benchmarks in this thread, but note that this is random accesses instead of sequential. So I'd like to ask @SouthEndMusic if the outcome is any different on his application, since he added the current cached index mechanism. 

Right now I have only added this mechanism in the `LinearInterpolation` cache, but if this design is OK'd then I'll add it onto the other interpolants.

image
image
image

(Note that these benchmarks are with a different computer than the previous ones).

SouthEndMusic:
@DaniGlez I like the concept, although I cannot make any assumptions about the distribution of `t` in my application as it is user input without restrictions on this distribution (apart from being sorted). I do feel like `looks_quasilinear` condition should be stricter, like looking at the variance of the intervals between the `t` as @ChrisRackauckas suggested. Because now something like this passes your test:

fig

DaniGlez:
Yeah, there are indeed false positives with this criterion; I felt that having minimal overhead with a cheap criterion is worth some false positives (particularly when you can manually opt-out at construction time) but it's absolutely up for debate.

ChrisRackauckas:
For a switchable mode, doing `quasiregular = 1e-2` would make a variance check that the variance is below 1e-2 and makes quasiregular true/false, and then `quasiregular = true/false` would skip the computation and just use the assumption from the user. I think defaulting to the check is a good idea because most cases I see would be fairly regular and so the optimization should be good for "most cases", and with this you just need to find a reasonable variance cutoff for the default.

DaniGlez:
Alright, I made a new proposal, it's similar to the variance one but I take the different w.r.t. the straight line, which seems more directly relevant (you could build a contrived counterexample to straightforward `var(diff(t))` with sawtooth increments that would nevertheless benefit from linear guessing). Let me know what you think about this one.

ChrisRackauckas:
Seems like the right path to me, though `seems_linear` and `quasiregular` should be unified into a single naming. Other than that, this seems good.

DaniGlez:
Ah, yeah, the flag name slipped. I'll change it later and apply it to other interpolant caches if everyone is happy with this.

src/interpolation_caches.jl
```diff
@@ -22,9 +22,13 @@ struct LinearInterpolation{uType, tType, IType, pType, T} <: AbstractInterpolati
...
     safetycopy::Bool
-    function LinearInterpolation(u, t, I, p, extrapolate, safetycopy)
+    seems_linear::Bool
+    function LinearInterpolation(
+            u, t, I, p, extrapolate, safetycopy; quasiregular = 1e-2)
+        quasiregular = quasiregular isa Bool ? quasiregular :
+                       looks_quasiregular(t; threshold = quasiregular)
         new{typeof(u), typeof(t), typeof(I), typeof(p.slope), eltype(u)}(
-            u, t, I, p, extrapolate, Ref(1), safetycopy)
+            u, t, I, p, extrapolate, Ref(1), safetycopy, quasiregular)
```
ChrisRackauckas, line 31: seems_linear

src/interpolation_utils.jl
```diff
@@ -106,6 +106,22 @@ function munge_data(U::StridedMatrix, t::AbstractVector, safetycopy::Bool)
     return readonly_wrap(U), readonly_wrap(t)
 end
 
+# """
+# Determine if the abscissae are regularly distributed, taking the standard \
+# deviation of the difference between the array of abscissae with respect to \
+# the straight line linking its first and last elements
+# """
+function looks_quasiregular(t; threshold = 1e-2)
```
ChrisRackauckas, line 114:
```suggestion
function looks_linear(t; threshold = 1e-2)
```

src/interpolation_methods.jl
```diff
@@ -1,7 +1,14 @@
 function _interpolate(A, t)
     ((t < A.t[1] || t > A.t[end]) && !A.extrapolate) &&
         throw(ExtrapolationError())
-    val, idx_prev = _interpolate(A, t, A.idx_prev[])
+    idx_guess = if hasfield(typeof(A), :seems_linear) && A.seems_linear
```
SouthEndMusic, line 4: It might be nicer to have all indexing logic in one place, that way you don't have to repeat it for integrals and derivatives. Say you leave this part unchanged, and refactor `get_idx` with this signature:

 `get_idx(A::AbstractInterpolation, t, iguess; lb = 1, ub_shift = -1, idx_shift = 0, side = :last)`

here you can get `tvec` and `seems_linear` from `A`, and overrule the given `iguess` with your one under the desired conditions.

sathvikbhagavan: yeah, makes sense.

ChrisRackauckas: Yes, and the bigger thing I haven't brought up is, it would then be great to make the OrdinaryDiffEq/DiffEqBase interpolations make use it of.

One step at a time 😅

DaniGlez:
Alright, finally got back to finishing this with remaining caches and suggestions. The docstring component is maybe a bit too wordy but everything seems to be working fine.

src/interpolation_caches.jl
```diff
@@ -12,7 +12,12 @@ Extrapolation extends the last linear polynomial on each side.
 ## Keyword Arguments
 
   - `extrapolate`: boolean value to allow extrapolation. Defaults to `false`.
-  - `cache_parameters`: precompute parameters at initialization for faster interpolation computations. Note: if activated, `u` and `t` should not be modified. Defaults to `false`.
+  - `cache_parameters`: precompute parameters at initialization for faster interpolation
+    computations. Note: if activated, `u` and `t` should not be modified. Defaults to `false`.
+  - `assume_linear_t`: boolean value to specify a faster index lookup behaviour for
+    evenly-distributed abscissae. Alternatively, a numerical threshold may be specified
+    for a test based on the normalized standard deviation of the difference with respect
```
sathvikbhagavan, line 19: I think it would be good to refer to the docstring of the function `looks_linear` such that it is clear from users' perspective about how it is checked.

src/interpolation_caches.jl
```diff
@@ -12,7 +12,12 @@ Extrapolation extends the last linear polynomial on each side.
...
+    to the straight line. Defaults to 1e-2
```
sathvikbhagavan, line 20:
```suggestion
    to the straight line. Defaults to 1e-2.
```

Can you do the same thing for all?

src/interpolation_caches.jl
```diff
@@ -22,18 +27,22 @@ struct LinearInterpolation{uType, tType, IType, pType, T} <: AbstractInterpolati
     extrapolate::Bool
     idx_prev::Base.RefValue{Int}
     cache_parameters::Bool
-    function LinearInterpolation(u, t, I, p, extrapolate, cache_parameters)
+    use_linear_lookup::Bool
```
sathvikbhagavan, line 30:
```suggestion
    linear_lookup::Bool
```

It would be more consistent with how `linear_lookup` is computed in the constructor.

src/interpolation_utils.jl
```diff
@@ -94,7 +94,36 @@ function munge_data(U::StridedMatrix, t::AbstractVector)
     return U, t
 end
 
-function get_idx(tvec, t, iguess; lb = 1, ub_shift = -1, idx_shift = 0, side = :last)
+seems_linear(assume_linear_t::Bool, _) = assume_linear_t
+seems_linear(assume_linear_t::Number, t) = looks_linear(t; threshold = assume_linear_t)
+
+# """
```
sathvikbhagavan, line 100: Can you revive this docstring?

sathvikbhagavan (commented):
A couple of comments. Can you add a few tests which demonstrate the speedup mentioned (with regular spaced t) in comments in the PR to make sure everything is working correctly?

DaniGlez:
I've incorporated your suggestions, @sathvikbhagavan. I'm a bit unsure about the benchmark, because it's a bit expensive as a test (a few seconds) compared to most of the test suite and CPU timings are non-deterministic (thus, the test might fail because of some random stuff). 

I also have a question regarding the `looks_linear` docstring; I added it to the `@docs` in the manual to stop CI from complaining, but does that make it public API? In my view it should remain as an internal implementation-specific function.

docs/src/manual.md
```diff
@@ -13,4 +13,5 @@ BSplineApprox
 CubicHermiteSpline
 PCHIPInterpolation
 QuinticHermiteSpline
+looks_linear
```
sathvikbhagavan, line 16: Make this into a different docs block like: 

````
# Utility Functions

```@docs
DataInterpolations.looks_linear
```
````

As `looks_linear` is not exported, need to use module name.

File: src/interpolation_utils.jl [hunk 9]
```diff
@@ -94,7 +94,36 @@ function munge_data(U::StridedMatrix, t::AbstractVector)
...
 
-function get_idx(tvec, t, iguess; lb = 1, ub_shift = -1, idx_shift = 0, side = :last)
+seems_linear(assume_linear_t::Bool, _) = assume_linear_t
+seems_linear(assume_linear_t::Number, t) = looks_linear(t; threshold = assume_linear_t)
+
+"""
+Determine if the abscissae `t` are regularly distributed, taking the standard deviation of \
+the difference between the array of abscissae with respect to the straight line linking \
+its first and last elements, normalized by the range of `t`.
+"""
```
sathvikbhagavan, line 104:
```suggestion
"""
	looks_linear(t; threshold = 1e-2)

Determine if the abscissae `t` are regularly distributed, taking the standard deviation of
the difference between the array of abscissae with respect to the straight line linking
its first and last elements, normalized by the range of `t`. If it is less than `threshold`, it looks linear.
"""
```

sathvikbhagavan (commented):
A couple of small comments.
I had a quick question - should this not be used in Lagrange Interpolation or even if it is used, it wouldn't make much difference?
Last step, run the julia formatter.

DaniGlez:
> A couple of small comments. I had a quick question - should this not be used in Lagrange Interpolation or even if it is used, it wouldn't make much difference? Last step, run the julia formatter.

In Lagrangian interpolation, each datapoint contribuyes a term towards the result, thus there is no need to call `get_idx` - you always use all your data to compute the interpolated value.

docs/src/manual.md
DaniGlez: Thanks, will change it in the next commit

DaniGlez:
Not sure what's wrong with the format of `docs/src/manual.md`, running JuliaFormatter 1.0.59 on it does not make any changes.

sathvikbhagavan:
>Not sure what's wrong with the format of docs/src/manual.md, running JuliaFormatter 1.0.59 on it does not make any changes.

Add a newline at the end of file.

DaniGlez:
I had thought that would not be an issue since the live version https://github.com/SciML/DataInterpolations.jl/blob/master/docs/src/manual.md?plain=1 seems to not have it either, let's see now

ChrisRackauckas (approved):

ChrisRackauckas:
Amazing, thanks for seeing this through. 

It would be great to see if we could reuse this as well with the diffeq interpolations, but that can wait.

ChrisRackauckas:
I mean, for the special interpolations for the ODE solvers, there is a similar search phase to find where to interpolate, and that just uses a naive findfirst. There's now a lot of optimizations used here to find the right time more quickly, and it would be nice to get those used there as well.

SouthEndMusic:
Maybe move the functionality to `FindFirstFunctions`?

ChrisRackauckas:
Possibly yeah, since this is really a smarter findfirst.

sathvikbhagavan:
yeah, abstracting this in findfirst would be better.

PR #209, opened 2023-12-04
typos CI

ArnoStrouwen:


sathvikbhagavan:
Does this check typos for docs as well?

ArnoStrouwen:
Yes. It is just less likely to find them there, since I already went over those manually before.

ChrisRackauckas:
What's the reason this one didn't get Aqua?

ChrisRackauckas:
What package uses TreeViews? I thought it was removed awhile back?

ArnoStrouwen:
I think it is https://github.com/SciML/MultiScaleArrays.jl/pull/83
and I also got the same error at Symbolics, but I did not confirm it was TreeViews that caused it.

ChrisRackauckas:
We should remove all overloads of that. It only applies to Juno/Atom which is long gone now.

ArnoStrouwen:
https://github.com/JuliaSymbolics/Symbolics.jl/blob/master/Project.toml#L39

PR #241, opened 2024-04-16
Add missing Symbolics.wrap

lassepe:


sathvikbhagavan (commented):
Can you add a test which basically caused it to fail?

Issue #240, opened 2024-04-12
Data interpolation unexpectedly behaves like a vector


gaurav-arya:
The behaviour is defined here:

https://github.com/SciML/DataInterpolations.jl/blob/0641b6929f564ca0ded6ea50015e8ac02ee79861/src/DataInterpolations.jl#L5-L16

This can lead to surprising results beyond just the getindex. For example, running `copy` on a data interpolation produces a vector with the u's and t's concatenated.

ChrisRackauckas:
It was due to how it would interact with ModelingToolkit in the past for controls applications. There are many instances where you want to treat the interpolation as a vector of parameters to fit. But yeah the ecosystem grew around this and now there's better ways of specifying a control function to fit using MTK and this is an interface oddity that remains. Probably should be removed.

PR #247, opened 2024-05-06
refactor!: remove indexing dispatches and add dispatch for higher order derivatives with Symbolics

sathvikbhagavan:
If `AbstractVector` subtyping is not removed, We get errors like:

```julia
julia> df2 = expand_derivatives(D2(f(τ)))
ERROR: Differentiation with array expressions is not yet supported
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] occursin_info
    @ ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:67 [inlined]
  [3] (::Symbolics.var"#254#256"{SymbolicUtils.BasicSymbolic{Real}, SymbolicUtils.BasicSymbolic{Real}})(a::LinearInterpolation{Vector{Float64}, Vector{Float64}, true, Float64})
    @ Symbolics ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:100
  [4] iterate
    @ ./generator.jl:47 [inlined]
  [5] _collect(c::Vector{Any}, itr::Base.Generator{Vector{…}, Symbolics.var"#254#256"{…}}, ::Base.EltypeUnknown, isz::Base.HasShape{1})
    @ Base ./array.jl:854
  [6] collect_similar
    @ ./array.jl:763 [inlined]
  [7] map
    @ ./abstractarray.jl:3285 [inlined]
  [8] occursin_info(x::SymbolicUtils.BasicSymbolic{Real}, expr::SymbolicUtils.BasicSymbolic{Real}, fail::Bool)
    @ Symbolics ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:100
  [9] occursin_info(x::SymbolicUtils.BasicSymbolic{Real}, expr::SymbolicUtils.BasicSymbolic{Real})
    @ Symbolics ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:65
 [10] expand_derivatives(O::SymbolicUtils.BasicSymbolic{Real}, simplify::Bool; occurrences::Nothing)
    @ Symbolics ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:177
 [11] expand_derivatives
    @ ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:171 [inlined]
 [12] expand_derivatives(O::SymbolicUtils.BasicSymbolic{Real})
    @ Symbolics ~/.julia/packages/Symbolics/Eas9m/src/diff.jl:171
 [13] top-level scope
    @ REPL[12]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

EDIT: This is indeed breaking.

Also fixes: #240

### DIFF
--- a/ext/DataInterpolationsSymbolicsExt.jl
+++ b/ext/DataInterpolationsSymbolicsExt.jl
@@ -21,6 +21,10 @@ function derivative(interp::AbstractInterpolation, t::Num, order = 1)
 end
 SymbolicUtils.promote_symtype(::typeof(derivative), _...) = Real
 
+function Symbolics.derivative(::typeof(derivative), args::NTuple{3, Any}, ::Val{2})
+    Symbolics.unwrap(derivative(args[1], Symbolics.wrap(args[2]), args[3] + 1))
+end
+
 function Symbolics.derivative(interp::AbstractInterpolation, args::NTuple{1, Any}, ::Val{1})
     Symbolics.unwrap(derivative(interp, Symbolics.wrap(args[1])))
 end
--- a/ext/DataInterpolationsOptimExt.jl
+++ b/ext/DataInterpolationsOptimExt.jl
@@ -31,7 +31,7 @@ function Curvefit(u,
         mfit = optimize(od, lb, ub, p0, Fminbox(alg))
     end
     pmin = Optim.minimizer(mfit)
-    CurvefitCache{true}(u, t, model, p0, ub, lb, alg, pmin, extrapolate)
+    CurvefitCache(u, t, model, p0, ub, lb, alg, pmin, extrapolate)
 end
 
 # Curvefit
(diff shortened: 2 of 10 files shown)


ext/DataInterpolationsSymbolicsExt.jl
```diff
@@ -21,6 +21,10 @@ function derivative(interp::AbstractInterpolation, t::Num, order = 1)
 end
 SymbolicUtils.promote_symtype(::typeof(derivative), _...) = Real
 
+function Symbolics.derivative(::typeof(derivative), args::NTuple{3, Any}, ::Val{2})
+    Symbolics.unwrap(derivative(args[1], Symbolics.wrap(args[2]), args[3] + 1))
+end
```
ChrisRackauckas, line 26: Order is an integer, how is this derivative defined?

ChrisRackauckas:
It is breaking but nothing is relying on it. Let's do it. Remove the indexing overloads as well, and the type parameter associated with it. I.e. Fix https://github.com/SciML/DataInterpolations.jl/issues/240

ChrisRackauckas (approved):

PR #250, opened 2024-05-07
chore: format NEWS.md and bump DataInterpolations version in docs

sathvikbhagavan:
followup of #249


sathvikbhagavan:
Only codecov failures

Issue #253, opened 2024-05-30
Type instability: Interpolations in v5 return Vector{Any} with all-Vector{Float64} inputs


moble:
**Describe the example**

With the latest version of this package, the various interpolations like `CubicSpline` returns `Vector{Any}` even when all inputs are `Vector{Float64}`.  This doesn't happen in v4 of this package.

**Minimal Reproducible Example 👇**

This is what I get in a temp environment using `add DataInterpolations@4`:
```julia
julia> using DataInterpolations

julia> t = Float64[0.1, 0.2, 0.3, 0.4, 0.5]; CubicSpline(t, t)(t)
5-element Vector{Float64}:
 0.1
 0.2
 0.3
 0.4
 0.5
```
Note that the output is a `Vector{Float64}`.  I get the same results with all the interpolations (except `BSplineInterpolation` which I haven't tried).

**Error & Stacktrace ⚠️**

Now in a temp environment using `add DataInterpolations@5`:
```julia
julia> using DataInterpolations

julia> t = Float64[0.1, 0.2, 0.3, 0.4, 0.5]; CubicSpline(t, t)(t)
5-element Vector{Any}:
 0.1
 0.2
 0.3
 0.4
 0.5
```
Note that the output is a `Vector{Any}`.  I get the same results with all the interpolations (except `BSplineInterpolation` which I haven't tried).

**Not Working Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia
Status `/private/var/folders/bh/rg4bdsl965l0572z3vtmwkr40000gn/T/jl_HUAoT0/Project.toml`
  [82cc6244] DataInterpolations v5.0.0
```

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

```julia
Status `/private/var/folders/bh/rg4bdsl965l0572z3vtmwkr40000gn/T/jl_HUAoT0/Manifest.toml`
  [bbf7d656] CommonSubexpressions v0.3.0
  [a8cc5b0e] Crayons v4.1.1
  [9a962f9c] DataAPI v1.16.0
  [82cc6244] DataInterpolations v5.0.0
  [e2d170a0] DataValueInterfaces v1.0.0
  [163ba53b] DiffResults v1.1.0
  [b552c78f] DiffRules v1.15.1
  [ffbed154] DocStringExtensions v0.9.3
  [64ca27bc] FindFirstFunctions v1.2.0
[...]

PR #254, opened 2024-05-31
fix: get element type from `interp.u` instead of  `interp`

sathvikbhagavan:
fixes: #253

In DataInterpolations@v4, `eltype(interp)` returned `eltype(interp.u)` as it subtyped `AbstractVector` of type which is the eltype of `interp.u`

In DataInterpolations@v5, `eltype(interp)` returns `Any` as the subtyping to AbstractVector is removed, hence the issue. Fixing it to use `eltype(interp.u)`.

Added a few tests to ensure the returned vectors have the same type as `interp.u`.


sathvikbhagavan:
Not sure why Symbolics.jl is failing to precompile - https://github.com/SciML/DataInterpolations.jl/actions/runs/9316593135/job/25645177538?pr=254#step:6:376

ChrisRackauckas:
Depwarns are turned to errors. I'll handle this today.

Issue #271, opened 2024-06-24
Increasing performance by caching


SouthEndMusic:
For my application I am looking at several features to improve the performance of interpolations:

- Caching: e.g. precompute and store the slopes of all segments for a linear interpolation
- Using the previous index as the `iguess` for the next evaluation; this is especially effective for my application where successive evaluations are close together

I have implemented these features in a downstream package: https://github.com/SouthEndMusic/CachedInterpolations.jl (recently rebranded from `SmoothInterpolation.jl`, new name not registered yet)

Would you be interested in implementing these features in `DataInterpolations.jl`?

ChrisRackauckas:
Yeah it would need benchmarks to go along with it but if it's hitting the performance then we'd definitely accept it.

SouthEndMusic:
Regarding caching: do you want to support data modification after initialization of the interpolation object?

ChrisRackauckas:
I'm not entirely sure what you're referring to.

SouthEndMusic:
Currently there is a single source of truth for the data `u,t` and so something like this gives the expected result:

```julia
u = [1.0, 1.0, 1.0]
t = [1.0, 2.0, 3.0]

itp = LinearInterpolation(u, t)

u[2] = 2.0

itp(1.5) # 1.5
```

However, if you cache the slopes at the initialisation of `itp` this will no longer work.

SouthEndMusic:
The `u, t` fields  can be wrapped by `ReadOnlyArray` from https://github.com/JuliaArrays/ReadOnlyArrays.jl so that they are at least not mutable when accessed as a field of the interpolation object.

ChrisRackauckas:
That is a partial solution, but your example would still be an issue.

SouthEndMusic:
The mutability of an existing object seems pretty immutable, better create a copy of `u, t` which are then wrapped in `ReadOnlyArray` if we're allocating for the cache in the constructor anyway.

ChrisRackauckas:
That would be a nice way to do it. And we can have `safetycopy = Val(true)` for whether to enable/disable this extra copy.

SouthEndMusic:
By the way, only this method of `munge_data` is not already copying `u, t`:

https://github.com/SouthEndMusic/DataInterpolations.jl/blob/088a9629a73e5ed52ed211422b227e159e8d11de/src/interpolation_utils.jl#L36-L38

SouthEndMusic:
[...]

PR #274, opened 2024-06-26
Cache parameters

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/271 (the other half)

## Checklist


SouthEndMusic:
That was a bit fiddly to get right (or at least pass the tests). Quick benchmark:

```julia
using DataInterpolations
using BenchmarkTools

N = 1000
u = rand(N)
t = cumsum(rand(N))

itp = LinearInterpolation(u, t)

t_eval = range(first(t), last(t), length = 10000)

@btime itp($t_eval)

# old: 162.200 μs (3 allocations: 78.23 KiB)
# new: 119.900 μs (3 allocations: 78.23 KiB)
```

SouthEndMusic:
Another benchmark:

```julia
using DataInterpolations
using BenchmarkTools

N = 1000
u = rand(N)
t = cumsum(rand(N))

itp = QuadraticInterpolation(u, t)

@btime DataInterpolations.integral(itp, t[end])

# old: 18.400 μs (3 allocations: 48 bytes)
# new:  9.000 μs (3 allocations: 48 bytes)
```

SouthEndMusic:
I'll leave it at caching parameters for `LinearInterpolation` and `QuadraticInterpolation` for this PR. 

An interesting one to cache parameters for next are `BsplineInterpolation` and `BsplineApprox`, as they now allocate at every  call in `spline_coefficients`:

```julia
using DataInterpolations
using BenchmarkTools

N = 1000
u = rand(N)
t = cumsum(rand(N))

t = [0, 62.25, 109.66, 162.66, 205.8, 252.3]
u = [14.7, 11.51, 10.41, 14.95, 12.24, 11.22]
A1 = QuadraticInterpolation(u,t)
A2 = BSplineInterpolation(u, t, 2, :Uniform, :Uniform)
A3 = BSplineApprox(u,t, 2, 4, :Uniform, :Uniform)

ts = collect(range(first(t), last(t), length = 10000))

@btime A1.(ts)
# 91.100 μs (3 allocations: 78.25 KiB)

@btime A2.(ts)
# 394.500 μs (10003 allocations: 1.14 MiB)

@btime A3.(ts)
# 387.400 μs (10003 allocations: 1015.77 KiB)
```

SouthEndMusic:
Refactor of `LagrangeInterpolation` to be non-allocating and using `idx_prev`:

```julia
using DataInterpolations
using BenchmarkTools

N = 10
t = cumsum(rand(N))
u = rand(N)
itp = LagrangeInterpolation(u, t, N-1)

ts = range(first(t), last(t), length = 10000)

@btime itp($ts)
# old: 1.806 ms (20003 allocations: 2.82 MiB)
# new: 465.300 μs (3 allocations: 78.23 KiB)
```

SouthEndMusic:
Using knowledge of which spline coefficients are non-zero:

```julia
using DataInterpolations
using BenchmarkTools

t = cumsum(rand(100))
u = rand(100)
A = BSplineInterpolation(u, t, 2, :Uniform, :Uniform)

ts = collect(range(first(t), last(t), length = 10000))

@btime A($ts)
# old: 1.161 ms (2 allocations: 78.17 KiB)
# new: 663.900 μs (2 allocations: 78.17 KiB)

```

src/online.jl
```diff
@@ -1,40 +1,63 @@
 import Base: append!, push!
 
-function push!(di::LinearInterpolation{U, T}, u::eltype(U), t::eltype(T)) where {U, T}
-    push!(di.u, u)
-    push!(di.t, t)
-    di
+function push!(A::LinearInterpolation{U, T}, u::eltype(U), t::eltype(T)) where {U, T}
```
sathvikbhagavan, line 3: Shouldn't this be allowed only if `safetycopy` is false?

test/interpolation_tests.jl
```diff
@@ -75,14 +75,6 @@ end
     @test A(3.0) == 2.0
     @test A(4.0) == 3.0
 
-    u = [0.0, NaN, 2.0, 3.0]
```
sathvikbhagavan, line 78: Why was this removed?

sathvikbhagavan (commented):
Looks good! Can you also add a few tests to ensure all cache parameters are set properly?

test/interpolation_tests.jl
SouthEndMusic: Unless I am missing something very subtle, this block of tests is identical to the previous one

src/online.jl
SouthEndMusic: I don't think I fully understand the reasoning here. If `safetycopy` is true, then the `u, t` vectors within the interpolation object can safely be assumed to be only used within the context of this package. Surely then it is safe to add data to it? Only if `u, t` are also used in code outside the `DataInterpolations` context do you run the risk of breaking that code by modifying `u, t`.

test/interpolation_tests.jl
sathvikbhagavan: Ah yes, didn't see that.

src/online.jl
sathvikbhagavan: I think I understand your point. It comes down what should be the semantics of `push!` for interpolation objects. Currently it mutates the original array. Maybe we should handle both cases? If it is true, it would push to the copy, if it is false it would mutate the original array. For both we have to compute the parameters again. 
@ChrisRackauckas, your thoughts?

docs/src/interface.md
```diff
@@ -35,6 +35,23 @@ A2(300.0)
     
     The values computed beyond the range of the time points provided during interpolation will not be reliable, as these methods only perform well within the range and the first/last piece polynomial fit is extrapolated on either side which might not reflect the true nature of the data.
 
+The keyword `safetycopy=false` can be passed to make sure no copies of `u` and `t` are made when initializing the interpolation object.
```
sathvikbhagavan, line 38:
```suggestion
The keyword `safetycopy = false` can be passed to make sure no copies of `u` and `t` are made when initializing the interpolation object.
```

docs/src/interface.md
```diff
@@ -35,6 +35,23 @@ A2(300.0)
...
+
+```@example interface
+A3 = QuadraticInterpolation(u, t; safetycopy = false)
+
+# Check for same memory
+u === A3.u.parent
+```
+
+Note that this does not prevent allocation in every interpolation constructor call, because parameter values are cached for some interpolation types.
```
sathvikbhagavan, line 47: Can you also add a line describing which interpolations support caching?

src/online.jl
```diff
@@ -1,40 +1,69 @@
...
+    !A.safetycopy && throw(UnsafeAddDataError())
```
sathvikbhagavan, line 4: Is this needed? Shouldn't we allow for both?

test/integral_tests.jl
```diff
@@ -83,7 +83,7 @@ end
 
 @testset "LagrangeInterpolation" begin
     u = [1.0, 4.0, 9.0]
-    t = [1.0, 2.0, 3.0]
+    t = [1.0, 2.0, 6.0]
```
sathvikbhagavan, line 86: Why was this changed?

test/online_tests.jl
```diff
@@ -1,18 +1,31 @@
 using DataInterpolations
 
-t = [1, 2, 3]
-u = [0, 1, 0]
+t1 = Float64[1, 2, 3]
+u1 = Float64[0, 1, 0]
```
sathvikbhagavan, line 4:
```suggestion
t1 = [1.0, 2.0, 3.0]
u1 = [0.0, 1.0, 0.0]
```

test/online_tests.jl
```diff
@@ -1,18 +1,31 @@
...
 
-for di in [LinearInterpolation, QuadraticInterpolation, ConstantInterpolation]
-    li = di(copy(u), copy(t))
-    append!(li, u, t)
-    li2 = di(vcat(u, u), vcat(t, t))
-    @test li.u == li2.u
-    @test li.t == li2.t
+t2 = Float64[4, 5, 6]
+u2 = Float64[1, 2, 1]
```
sathvikbhagavan, line 7:
```suggestion
t2 = [4.0, 5.0, 6.0]
u2 = [1.0, 2.0, 1.0]
```

test/integral_tests.jl
SouthEndMusic: I added this function to conveniently test whether the implementation of a new interpolation type is complete: https://github.com/SciML/DataInterpolations.jl/pull/274/files#diff-f66955aae2aee262495ff068999fe35f8501cd1cbef7b1057aa26d2f1099a4f3R6-R17

To do this, I moved throwing an error when integration is not defined from specialised methods of `integral` to methods of `_integral`. However, that now means that the extrapolation error is thrown before the 'integral not found' error. So, to still test the integral not found error, I had to make sure that the test for this does not try to extrapolate.

src/online.jl
SouthEndMusic: That is fine by me, I think it is clear enough that this leads to mutation of the input `u, t` if someone explicitly sets `safetycopy = false`.

src/interpolation_utils.jl
```diff
@@ -1,14 +1,42 @@
...
+            A.idxs[i] = idx_max
+        end
+    end
+    return idx
 end
 
-function spline_coefficients(n, d, k, u::Number)
-    N = zeros(eltype(u), n)
+function spline_coefficients!(N, d, k, u::Number)
+    N .= 0
```
sathvikbhagavan, line 33:
```suggestion
    N .= zero(u)
```

docs/src/interface.md
```diff
@@ -35,6 +35,23 @@ A2(300.0)
...
+```@example interface
+A3 = QuadraticInterpolation(u, t; safetycopy = false)
+
+# Check for same memory
+u === A3.u.parent
+```
+
+Note that this does not prevent allocation in every interpolation constructor call, because parameter values are cached for all interpolation types except `ConstantInterpolation`.
+
+Because of the caching of parameters which depend on `u` and `t`, this data should not be mutated. Therefore `u` and `t` are wrapped in a `ReadOnlyArray`.
```
sathvikbhagavan, line 49:
```suggestion
Because of the caching of parameters which depend on `u` and `t`, this data should not be mutated. Therefore `u` and `t` are wrapped in a `ReadOnlyArray` from [ReadOnlyArrays.jl](https://github.com/JuliaArrays/ReadOnlyArrays.jl).
```

sathvikbhagavan (commented):
Almost there. Just a few cleanups.

sathvikbhagavan (approved):
LGTM! Nice work! Thanks!

ChrisRackauckas (approved):

PR #288, opened 2024-07-06
refactor: use BaryCentric Formula for Lagrange Interpolation

### LINKED ISSUE #75: Use barycentric formula for Lagrange interpolation
It looks like you are using the textbook Lagrange interpolation formula. Trefethen has some articles on why this is numerically unstable and you should use the alternative barycentric formula instead. (This evaluates exactly the same interpolating polynomial in a different way.)

sathvikbhagavan:
fixes: #75

Derivatives still need some fixing.


sathvikbhagavan:
superseded by #596

PR #315, opened 2024-07-27
Refactor parameter caching + add Zygote tests

### LINKED ISSUE #289: Improving autodiff performance
I need to differentiate through the `DataInterpolations.jl` wrt to the new evaluation points and the input data.
Here a MWE
```
n = 64
x = vcat([0.], sort(rand(n-2)), [1.])
x1 = vcat([0.], sort(rand(n-2)), [1.])
y = rand(n);

function di_spline(y,x,xn)
    spline = QuadraticSpline(y,x, extrapolate = true)
    return spline.(xn)
end
```
Although computing such an interpolation is very efficient
```
@benchmark sum(di_spline($y,$x,$x1))
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.917 μs … 336.010 μs  ┊ GC (min … max): 0.00% … 98.41%
 Time  (median):     2.036 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.212 μs ±   4.597 μs  ┊ GC (mean ± σ):  3.49% ±  2.40%

  ▂▆██▇▆▅▄▂▂▆▆▅▂▁              ▂▁▁                            ▂
  █████████████████▇▇████▆▇▇▆███████▆▅▅▆▅▅▅▅▅▆▄▆▄▄▄▆▆▆▇▆▆▆▆▅▆ █
  1.92 μs      Histogram: log(frequency) by time      3.46 μs <

 Memory estimate: 3.42 KiB, allocs estimate: 7.
```
The computation of the gradient is three orders of magnitudes slower
```
@benchmark gradient($y->sum(di_spline($y,$x,$x1)), $y)
BenchmarkTools.Trial: 1288 samples with 1 evaluation.
 Range (min … max):  3.580 ms …   7.025 ms  ┊ GC (min … max): 0.00% … 44.98%
 Time  (median):     3.727 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.880 ms ± 565.918 μs  ┊ GC (mean ± σ):  3.12% ±  8.32%

  ▆██▇▆▅▄▁▁▁                                                   
  ██████████▇█▆▅▅▄▁▅▅▄▄▁▅▁▁▄▁▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▇▅▆▇▄▅▅▄▅▆▄▅▇▅▇ █
  3.58 ms      Histogram: log(frequency) by time      6.45 ms <

 Memory estimate: 2.31 MiB, allocs estimate: 38892.

```

As a solution, I think proper adjoints need to be added to the library. 
I have opened a discourse thread [here](https://discourse.julialang.org/t/datainterpolations-jl-and-zygote-jl/113208/15) and given @ChrisRackauckas answer, I opened this issue.
The main question is: assuming there is some interest in adding this feature, how to do it? In the aforementioned thread, I did it rewriting the whole function in `DataInterpolations`, dividing it in smaller functions and writing the adjoint for each of them. How would you like to proceed? Writing the constructor adjoint should not be an issue, I am more concerned about writing the adjoint for the interpolation
```
function _interpolate(A::QuadraticSpline{<:AbstractVector}, t::Number, iguess)
    idx = get_idx(A.t, t, iguess; lb = 2, ub_shift = 0, side = :first)
    Cᵢ = A.u[idx - 1]
    σ = 1 // 2 * (A.z[idx] - A.z[idx - 1]) / (A.t[idx] - A.t[idx - 1])
    return A.z[idx - 1] * (t - A.t[idx - 1]) + σ * (t - A.t[idx - 1])^2 + Cᵢ, idx
end
```

If anyone is willing to help me with one case at least (this one with `QuadraticSpline`, for instance) I will implement the same for other cases.
Thank you in advance!

### LINKED ISSUE #298: Zygote support broke in new release
See https://github.com/LuxDL/LuxLib.jl/actions/runs/9965073839/job/27534650459?pr=86#step:6:2929. Started showing up on update.

Probably needs a @non_differentiable on `methods`

### LINKED ISSUE #301: Remove `ReadOnlyArray` from type parameters
I want to hardcode the concrete type of an interpolation object as a field of a struct. In my opinion it shouldn't be required to have `ReadOnlyArrays` as an explicit dependency of my package to do this. I start to like this wrapper less and less 🙃 

Maybe a good way forward is to get rid of this wrapper, make caching parameters opt-in and properly document that if you do this you should not update `u, t`. This has the advantage that the default setting is more compatible with AD packages. Then we should only make sure that AD packages do not try to differentiate w.r.t. cached parameters.

SouthEndMusic:
Fixes #298
Fixes #301 

#289 can be a follow-up. Gradients w.r.t. u for `QuadraticSpline` are not broken anymore, but they are slow as mentioned in the issue.

The refactor includes:
- Caching parameters is opt-in, so users need to make the active decision to cache parameters. It is documented that this is not compatible with modifying `u` and `t`. We could if desired allow this with a dedicated interface which updates the parameters automatically. When parameters are not cached, they are computed on demand with the same functions.
- The `ReadOnlyArray` wrappers are removed, which simplifies the type parameters. I also noticed an inconsistency where I forgot to do it for integral inverse interpolations in the first place
- Also `safetycopy` as an option is removed again 

Zygote:
- I added Zygote tests for differentiation w.r.t. the interpolation input and w.r.t. the `u` of the data points for where that is supported (again). 
- I've added some examples in DataInterpolationsChainRulesCoreExt.jl of how to speed up gradients w.r.t. u with analytical expressions. Speeding up more interpolation types this way could be a follow up


SouthEndMusic:
How can I fix the Zygote compat?

Project.toml
```diff
@@ -33,14 +32,14 @@ LinearAlgebra = "1.10"
...
 QuadGK = "2.9.1"
-ReadOnlyArrays = "0.2.0"
 RecipesBase = "1.3"
 Reexport = "1"
 RegularizationTools = "0.6"
 SafeTestsets = "0.1"
 StableRNGs = "1"
 Symbolics = "5.29"
 Test = "1"
+Zygote = "^0"
```
ChrisRackauckas, line 42:
```suggestion
Zygote = "0.6"
```

ChrisRackauckas:
You can only compat weakdeps or deps. So added as a weakdep if we want the compat.

SouthEndMusic:
I've added tests for I think everything that was compatible with Zygote before. I've added some examples in `DataInterpolationsChainRulesCoreExt.jl` of how to speed up gradients w.r.t. u with analytical expressions. Speeding up more interpolation types this way could be a follow up

Issue #322, opened 2024-07-31
Error when extrapolating to -Inf in v6.1.0


bclyons12:
**Describe the bug 🐞**

A ConstantInterpolation that had previously been working in older versions of DataInterpolations is now broken in v6.1, throwing an `InexactError: trunc(Int64, -Inf)` when trying to interpolate to `-Inf`. This doesn't happen for every such interpolation, but an example is given below.

**Expected behavior**

The ConstantInterpolation should just give the first array value, as has been done in past versions.

**Minimal Reproducible Example 👇**

```julia
import Pkg
Pkg.activate(temp=true)
Pkg.add(name="DataInterpolations", version="6.1")
import DataInterpolations
x = [0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2.0, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8, 2.9, 3.0, 3.1, 3.2, 3.3, 3.4, 3.5, 3.6, 3.7, 3.8, 3.9, 4.0, 4.1, 4.2, 4.3, 4.4, 4.5, 4.6, 4.7, 4.8, 4.9, 5.0, 5.1, 5.2, 5.3, 5.4, 5.5, 5.6, 5.7, 5.8, 5.9, 6.0, 6.1, 6.2, 6.3, 6.4, 6.5, 6.6, 6.7, 6.8, 6.9, 7.0, 7.1, 7.2, 7.3, 7.4, 7.5, 7.6, 7.7, 7.8, 7.9, 8.0, 8.1, 8.2, 8.3, 8.4, 8.5, 8.6, 8.7, 8.8, 8.9, 9.0, 9.1, 9.2, 9.3, 9.4, 9.5, 9.6, 9.7, 9.8, 9.9]
[...]

PR #323, opened 2024-08-01
Use index guessing functionality moved to `FindFirstFunctions.jl` (`Guesser`)

SouthEndMusic:
Fix https://github.com/SciML/DataInterpolations.jl/issues/322.


SouthEndMusic:
Update: will use the updated version of `FindFirstFunctions` which includes the fix

src/derivatives.jl
```diff
@@ -1,16 +1,11 @@
-function derivative(A, t, order = 1)
+function derivative(A, t, order = 1; iguess = A.iguesser)
```
sathvikbhagavan, line 1: why do we need `iguess` as a kwarg here?

SouthEndMusic: To support a user-supplied `iguess`:

https://github.com/SouthEndMusic/DataInterpolations.jl/blob/3debd2909676a0c315ee41a76b90fe9ab146d6e8/src/DataInterpolations.jl#L25-L27

SouthEndMusic: Ah wait no that's only for interpolation, shouldn't we support that for derivatives too like this?

SouthEndMusic: But in that case it should be documented

sathvikbhagavan: As the time steps is the same, we should just use iguesser from interpolation objects.

SouthEndMusic: I mean that we could support an integer iguess from the user that overrules `iguesser` because they have a very fast way to determine it, or do you not think that that is worth it?

sathvikbhagavan: I don't think that would be worth it. IMO, users should not give `iguess` themselves for each derivative/interpolation computation (not the construction).

SouthEndMusic: Ok, but you want to keep https://github.com/SouthEndMusic/DataInterpolations.jl/blob/3debd2909676a0c315ee41a76b90fe9ab146d6e8/src/DataInterpolations.jl#L25-L27 (as it is already there)?

sathvikbhagavan: I think we can remove it as it is not documented.

SouthEndMusic: Ok, thanks for taking the time to discuss this 👍

SouthEndMusic: Turns out it is not tested either

sathvikbhagavan (approved):
LGTM! Weirdly doc builds' registry is not getting updated and hence its failing.

ChrisRackauckas:
@thazhemadam is there some issue with the cache?

I'm going to just merge.

thazhemadam:
I think it might be a transient issue. I don't see it happening on `master`.

PR #324, opened 2024-08-06
Fix Zygote AD with `u` a matrix or vector of arrays

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/298 (for matrices and vectors of arrays)


ChrisRackauckas:
Test failure.

Issue #328, opened 2024-08-23
Indicate output dimensionality in `AbstractInterpolation` type


adrhill:
As a developer writing a package extension for DataInterpolations.jl (https://github.com/adrhill/SparseConnectivityTracer.jl/pull/178), it would be useful if all `AbstractInterpolation`s indicated their output dimensionality, e.g.

```Julia
abstract type AbstractInterpolation{T,N} end 
```

or `{N,T}` instead of just `{T}`:

https://github.com/SciML/DataInterpolations.jl/blob/536cb0243038af495656ab11a219724acd56a293/src/DataInterpolations.jl#L5

ChrisRackauckas:
That's probably fine.

sathvikbhagavan:
Yeah, that can be done. Will work on it soon.

PR #342, opened 2024-10-07
refactor: add output dimensionality as a parameter in AbstractInterpolations

sathvikbhagavan:
fixes: #328

### DIFF
--- a/src/interpolation_utils.jl
+++ b/src/interpolation_utils.jl
@@ -59,6 +59,19 @@ function spline_coefficients!(N, d, k, u::AbstractVector)
     return nothing
 end
 
+# Get Output Dimension for parameterizing AbstractInterpolations
+function get_output_dim(u::AbstractVector{<:Number})
+    return (1,)
+end
+
+function get_output_dim(u::AbstractVector)
+    return (length(first(u)),)
+end
+
+function get_output_dim(u::AbstractArray)
+    return size(u)[1:(end - 1)]
+end
+
 # helper function for data manipulation
 function munge_data(u::AbstractVector{<:Real}, t::AbstractVector{<:Real})
     return u, t
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -21,7 +21,7 @@ jobs:
         group:
           - Core
         version:
-          - '1'
+          - '1.10'
         os:
           - ubuntu-latest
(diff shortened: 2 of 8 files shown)


Project.toml
```diff
@@ -42,7 +42,7 @@ StableRNGs = "1"
 Symbolics = "5.29, 6"
 Test = "1"
 Zygote = "0.6.70"
-julia = "1.10"
+julia = "1.11"
```
ChrisRackauckas, line 45:
```suggestion
julia = "1.10"
```

no we stay on lts

Project.toml
```diff
@@ -30,7 +30,7 @@ ChainRulesCore = "1.24"
 FindFirstFunctions = "1.3"
 FiniteDifferences = "0.12.31"
 ForwardDiff = "0.10.36"
-LinearAlgebra = "1.10"
+LinearAlgebra = "1.11"
```
ChrisRackauckas, line 33: why is this required?

Project.toml
sathvikbhagavan: Alright, I have reverted the commit.

Project.toml
sathvikbhagavan: I have reverted the commit.

src/interpolation_utils.jl
```diff
@@ -59,6 +59,19 @@ function spline_coefficients!(N, d, k, u::AbstractVector)
...
+    return (1,)
+end
+
+function get_output_dim(u::AbstractVector)
+    return (length(first(u)),)
+end
+
+function get_output_dim(u::AbstractArray)
+    return size(u)[1:(end - 1)]
+end
```
devmotion, line 73: @sathvikbhagavan @ChrisRackauckas This breaks type inference of the constructors which causes problems in my downstream package (see #389). One way around this would be to allow (require?) users to specify the dimension as a type parameter when constructing the interpolation method.

However, it's not clear to me why this type parameter should be included at all. Arguably it also doesn't make sense at all for nonnumeric types (that I actually encounter when working with `ConstantInterpolation`). Would it be OK to revert this change?

If the output dimension is actually of interest to someone then it seems easy enough for them to use the logic of `get_output_dim` directly. Possibly this could also be made official by adding something like `output_dim` to the API (if actually necessary). This could also e.g. error for non-numeric types.

Issue #206, opened 2023-11-16
Inconsistency in Interface


sathvikbhagavan:
All methods can take in a Vector of numbers, few methods take in Matrices, Vector of Vectors etc. We should have a consistent interface for all or document this API, reason for it etc.

## Methods which can take in Matrices:

1. LinearInterpolation - https://github.com/SciML/DataInterpolations.jl/blob/master/src/interpolation_methods.jl#L27C26-L27C45
2. QuadraticInterpolation - https://github.com/SciML/DataInterpolations.jl/blob/master/src/interpolation_methods.jl#L50C26-L50C48
3. LagrangeInterpolation -  https://github.com/SciML/DataInterpolations.jl/blob/master/src/interpolation_methods.jl#L88C26-L88C47
4. ConstantInterpolation - https://github.com/SciML/DataInterpolations.jl/blob/master/src/interpolation_methods.jl#L146C26-L146C47

So, `AkimaInterpolation`, `QuadraticSpline`, `CubicSpline`, `BSplineInterpolation`, `BSplineApprox` does not work with Matrices

## Methods which does not work with Vector of Vectors:

`AkimaInterpolation`, `BSplineInterpolation`, `BSplineApprox`

bertini97:
Since a lot of SciML components make use of `RecursiveArrayTools`, it would be good to dispatch also on `AbstractVectorOfArray` and just act on the vector.

Also, `LinearInterpolation` does not work with vector of vectors:
```julia
MethodError: no method matching zero(::Type{Vector{Float64}})
```

sathvikbhagavan:
Now, all the methods are covered and the interface is consistent.

PR #343, opened 2024-10-07
CubicSplines: Add AbstractMatrix support

ashutosh-b-b:
## Checklist

### DIFF
--- a/src/interpolation_caches.jl
+++ b/src/interpolation_caches.jl
@@ -460,6 +460,37 @@ function CubicSpline(u::uType,
     CubicSpline(u, t, I, p, h[1:(n + 1)], z, extrapolate, cache_parameters, linear_lookup)
 end
 
+function CubicSpline(u::uType,
+    t;
+    extrapolate = false, cache_parameters = false,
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Why is this needed instead of the vector of vector form?

sathvikbhagavan:
I think we should have both forms. Its also part of fixing https://github.com/SciML/DataInterpolations.jl/issues/206

ashutosh-b-b:
> Why is this needed instead of the vector of vector form?

We can extend this to higher dimensional matrix. That is of course dependent on, whether we need that in DataInterpolations.jl

ChrisRackauckas:
I see, yes we should support fully then.

ashutosh-b-b:
Generalized this to AbstractArray and added tests for 3d arrays as well.

sathvikbhagavan:
Can you rebase on master?

sathvikbhagavan (approved):
LGTM! A rebase and should be good to merge!

ChrisRackauckas (approved):

PR #346, opened 2024-10-16
Intercept frule at the top level

YingboMa:


sathvikbhagavan:
For my understanding, why is this needed?

YingboMa:
Because symbolics stops tracing at the top most level.

PR #356, opened 2024-11-13
Extrapolation options

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/355.

This is what I have in mind:

```julia
using DataInterpolations
using Plots
using Random

Random.seed!(2)

t = cumsum(rand(5))
u = rand(5)

t_eval = range(-1, 3.5, length = 100)
p = scatter(t,u; label = "data")
title!("Extrapolation types")

for extrapolation_type in [:constant, :linear, :extension]
    A = QuadraticSpline(u,t; extrapolation_up = extrapolation_type, extrapolation_down = extrapolation_type)
    plot!(t_eval, A.(t_eval); label = string(extrapolation_type))
end

p
```
figure

Let me know whether this is a good approach before I apply it to all interpolation types, integrals, derivatives.

PR #359, opened 2024-11-16
Refactor integration and `QuadraticInterpolation`

SouthEndMusic:
@sathvikbhagavan I split this out from https://github.com/SciML/DataInterpolations.jl/pull/356 for ease of reviewing. I need the `QuadraticInteprolation` refactor for the integration refactor, and the integration refactor for https://github.com/SciML/DataInterpolations.jl/pull/356.


ext/DataInterpolationsChainRulesCoreExt.jl
```diff
@@ -74,6 +74,11 @@ function u_tangent(A::LinearInterpolation, t, Δ)
     out
 end
 
+function _quad_interp_indices(A::QuadraticInterpolation, t::Number, iguess)
```
sathvikbhagavan, line 77: Just to confirm, can't `u_tangent` for QuadraticInterpolation be reformulated using `α` and `β` parameters you added?

sathvikbhagavan (commented):
Left one comment. Are we getting any speedup from this implementation of QuadraticInterpolation than the earlier version? Otherwise, Looks good!

ext/DataInterpolationsChainRulesCoreExt.jl
SouthEndMusic: I'ts possible, but I suspect this is more efficient. Also, that might break AD types that do not support mutation when those parameters are cached?

SouthEndMusic: And you still need to have the logic somewhere to see which mode (forward/backward looking interpolation) was used for the `u` gradient.

sathvikbhagavan: Alright, then we can keep this.

sathvikbhagavan (approved):

ChrisRackauckas (approved):

Issue #360, opened 2024-11-16
Add extrapolation types `periodic` and `reflection`


SouthEndMusic:
As a follow up of https://github.com/SciML/DataInterpolations.jl/issues/355. Can be quite easily achieved with some simple transformations of `t`.

PR #363, opened 2024-11-22
Add extrapolation types `periodic` and `reflective`

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/360.

Follow-up of https://github.com/SciML/DataInterpolations.jl/pull/356.

```julia
using DataInterpolations
using Plots
using Random
Random.seed!(1)
u = rand(5)
t = cumsum(rand(5))
A = QuadraticSpline(u,t; extrapolation_left = ExtrapolationType.reflective, extrapolation_right = ExtrapolationType.periodic)
T = last(A.t) - first(A.t)
t_eval_left = range(first(t) - 2T, first(t), length = 250)
t_eval_right = range(last(t), last(t) + 2T, length = 250)
plot(A)
plot!(t_eval_left, A.(t_eval_left); label = "reflective")
plot!(t_eval_right, A.(t_eval_right); label = "periodic")
```

figure


SouthEndMusic:
Note: in this PR I add tests for integrating with cached integral data (which is cached when `cache_parameters = true`). I found that this had a bug which I fixed. The bug was possibly introduced by https://github.com/SciML/DataInterpolations.jl/pull/359. So I suggest not to make a new release until this PR is merged.

docs/src/extrapolation_methods.md
```diff
@@ -0,0 +1,90 @@
...
+plot(A)
+```
+
+Extrapolation behavior can be set left and right of the data simultaneously with the `extension` keyword, or left and right separately with the `extrapolation_left` and `extrapolation_right` keywords respectively.
+
+## `ExtrapolationType.none`
+
+This extrapolation type will throw an error when the input `t` is beyond the data in the specified direction.
+
+## `ExtrapolationType.constant`
```
ChrisRackauckas, line 22:
```suggestion
## `ExtrapolationType.None`

This extrapolation type will throw an error when the input `t` is beyond the data in the specified direction.

## `ExtrapolationType.Constant`
```

The other EnumX's have the caps so we should keep the same style.

ChrisRackauckas (approved):

ChrisRackauckas:
This is amazing!

PR #366, opened 2024-11-25
Make extrapolation with NaN Unitful-friendly

### LINKED ISSUE #365: Error rather than NaN result if Unitful arrays passed to LinearInterpolation
**Describe the bug 🐞**

Linear interpolation works as expected on arrays of Unitful numbers, but if a Unitful NaN is passed, an error is thrown rather than returning NaNs back out.

**Expected behavior**

Whatever 

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
interp = LinearInterpolation(rand(5), 1:5, extrapolate=true)
interp(NaN) # returns NaN
interp = LinearInterpolation(rand(5)*u"K", (1:5)u"s", extrapolate=true)
interp(NaN*u"s") # errors
```

**Error & Stacktrace ⚠️**

```julia
Stacktrace:
 [1] -(x::Quantity{Float64, 𝐓, Unitful.FreeUnits{…}}, y::Quantity{Float64, NoDims, Unitful.FreeUnits{…}})
   @ Unitful C:\Users\iwheeler\.julia\packages\Unitful\GYzMo\src\quantities.jl:137
 [2] -(x::Quantity{Float64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}, y::Int64)
   @ Base .\promotion.jl:424
 [3] _interpolate(A::LinearInterpolation{…}, t::Quantity{…}, iguess::FindFirstFunctions.Guesser{…})
   @ DataInterpolations C:\Users\iwheeler\.julia\packages\DataInterpolations\C8cNn\src\interpolation_methods.jl:22
 [4] _interpolate
   @ DataInterpolations C:\Users\iwheeler\.julia\packages\DataInterpolations\C8cNn\src\interpolation_methods.jl:4 [inlined]
 [5] (::LinearInterpolation{…})(t::Quantity{…})
   @ DataInterpolations C:\Users\iwheeler\.julia\packages\DataInterpolations\C8cNn\src\DataInterpolations.jl:24
 [6] top-level scope
   @ REPL[11]:1
```

**Additional context**

From the stacktrace, it looks to me like the NaN-handling here should be using `oneunit(...)` rather than `one(...)`
https://github.com/SciML/DataInterpolations.jl/blob/0cc058d33cc69859a0487c580d8cce8834a7dcc4/src/interpolation_methods.jl#L9-L15

Ickaser:
## Checklist

## Additional context

Would in theory close #365 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -41,6 +41,7 @@ SafeTestsets = "0.1"
 StableRNGs = "1"
 Symbolics = "5.29, 6"
 Test = "1"
+Unitful = "1"
 Zygote = "0.6.70"
 julia = "1.10"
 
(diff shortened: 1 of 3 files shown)


SouthEndMusic:
Maybe also add a test to prevent regression.

sathvikbhagavan:
Yes, its fine to add it as a test dependency

Ickaser:
OK, I've added a test for this specific case, and on my machine all the tests now pass. As far as I can tell, there were no similar NaN checks for other interpolation types, so this should be good to go I think.

ChrisRackauckas (approved):

Issue #364, opened 2024-11-22
Smooth equivalent of `ConstantInterpolation` in terms of integral


SouthEndMusic:
**Is your feature request related to a problem? Please describe.**

A bit of context: In my ODE problem for hydrological modelling, I want to model certain forcing events like precipitation. The precipitation is given as a time series, where constant interpolation ('forward fill') is assumed so that the exact total volume is known. Constant interpolation however does not play nice with ODE solvers because of its lack of smoothness. Therefore I want to have an interpolation type which:

- Is at least $C^1$ smooth;
- Has for each interval between data points the same integral as forward fill would have.

An additional requirement I have is that it is easy to extend this interpolation with new incoming data. 

**Describe the solution you’d like**

The naive solution of assuming constant interpolation on the first interval and then quadratic on the others to satisfy the above conditions is not stable:

```julia
using DataInterpolations
using CairoMakie
using Random
using FindFirstFunctions: searchsortedfirst
using QuadGK

fig = Figure()

ax_itp = Axis(fig[1,1]; title = "Interpolation")
ax_int = Axis(fig[2,1]; title = "Integral")

N = 6
Random.seed!(5)
u = rand(N)
t = cumsum(rand(N))
t_eval = range(first(t), last(t), length = 250)

A = ConstantInterpolation(u, t; cache_parameters = true)
lines!(ax_itp, t_eval, A.(t_eval); label = "ConstantInterpolation")
lines!(ax_int, t_eval, DataInterpolations.integral.(Ref(A), t_eval); label = "ConstantInterpolation")

a = zeros(N - 1)
b = zeros(N - 1)
c = zeros(N - 1)

# Assume constant interpolation on first interval
c[1] = u[1]

for i in 1:(N-2)
    Δt = t[i+1] - t[i]

    # value at t[i+1]
    u_ = @evalpoly Δt c[i] b[i] a[i]

    # derivative at t[i+1]
    du_ = @evalpoly Δt b[i] 2a[i]

    # integral over (t[i+1], t[i+2])
    Δt = t[i+2] - t[i+1]  
    I = Δt * u[i+1]

    # Next coefficients
    c[i+1] = u_
    b[i+1] = du_
    a[i+1] = 3 * (u[i+1] - u_ - (du_ * Δt) / 2) / Δt^2
end

function B(t_)
    i = clamp(searchsortedfirst(t, t_)-1, 1, N-1)
[...]

PR #367, opened 2024-11-25
Add `SmoothedConstantInterpolation` 

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/364.

## Checklist

## Additional context

```julia
using DataInterpolations
using Plots
using Random

N = 10
Random.seed!(8)
u = rand(N)
t = cumsum(rand(N))
A1 = ConstantInterpolation(u, t)

p = plot()
plot!(A1)

for d_max in 00.05:0.05:0.2
    A2 = SmoothedConstantInterpolation(u, t, cache_parameters = true; d_max)
    plot!(A2)
end

p
```

figure


SouthEndMusic:
Fun feature; if `SmoothedConstantInterpolation` is used with `extrapolation = ExtrapolationType.Periodic`, then the transitions are also smooth:

```julia
using DataInterpolations
using Plots
using Random

N = 4
Random.seed!(8)
u = rand(N)
t = cumsum(rand(N))
Δt = t[end] - t[1]
t_eval = range(first(t) - Δt, last(t) + Δt, length = 500)

A = SmoothedConstantInterpolation(u, t; extrapolation = ExtrapolationType.Periodic, d_max = 0.2)

plot(t_eval, A.(t_eval))
scatter!(t[1:end-1], u[1:end-1]; label = "data")
scatter!(t[1:end-1] .+ Δt, u[1:end-1]; label = "data one period forward")
scatter!(t[1:end-1] .- Δt, u[1:end-1]; label = "data one period back")
```

plot

src/interpolation_caches.jl
```diff
@@ -366,6 +366,78 @@ function ConstantInterpolation(
         cache_parameters, assume_linear_t)
 end
 
+"""
+    SmoothedConstantInterpolation(u, t; d_max = Inf, extrapolate = false,
+        cache_parameters = false, assume_linear_t = 1e-2)
+
+It is a method for interpolating constantly with forward fill, with smoothing around the
+value transitions to make the curve continuously differentiable while the integral never
+drifts far from the integral of constant interpolation. In this interpolation type u[end] is ignored.
```
sathvikbhagavan, line 375:
```suggestion
drifts far from the integral of constant interpolation. In this interpolation type, u[end] is ignored.
```

src/interpolation_caches.jl
```diff
@@ -366,6 +366,78 @@ function ConstantInterpolation(
...
+
+## Arguments
+
+  - `u`: data points.
+  - `t`: time points.
+
+## Keyword Arguments
+
+  - `d_max`: the maximum distance in `t` from the data points the smoothing is allowed to reach.
```
sathvikbhagavan, line 384: For users, this might not be very clear. Can you add an explanation like this image in https://github.com/SciML/DataInterpolations.jl/issues/364#issuecomment-2495610505

sathvikbhagavan (approved):

ChrisRackauckas:
Conflicts?

docs/src/methods.md
```diff
@@ -77,6 +77,24 @@ A = ConstantInterpolation(u, t, dir = :right)
...
+with the keyword argument `d_max`.
+
+```@example tutorial
+A = ConstantInterpolation(u, t)
+plot(A)
+A = SmoothedConstantInterpolation(u, t; d_max = 10)
+plot!(A)
+```
+
+Note that `u[end]` is ignored.
```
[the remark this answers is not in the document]
SouthEndMusic: Yes, also for extrapolation

SouthEndMusic:
@visr I added the function `get_transition_ts` to know which `tstops` to pass to the integrator so that it doesn't jump over the smooth transitions:

image

ChrisRackauckas:
What's the reason the end is ignored and it doesn't just constant extrapolate?

SouthEndMusic:
The key properties of this interpolation type are that it is differentiable everywhere and sufficiently close to constant interpolation. I achieve the diffentiability by replacing the stepwise transition in each data point by a differentiable transition in an interval around each data point.

What that means for extrapolation is a bit murky. I already made it so that the extrapolation is smooth in the case of periodic extrapolation, which means that the interpolation around the edges of the domain has to account for that. I'd argue that for constant extrapolation, ignoring the last data point is also best. Maybe for extension extrapolation, the last data point can be used, but that feels a bit inconsistent l.

sathvikbhagavan:
If I understand correctly, can't we do the same transition from second last to the last point after which it becomes constant? But this is strictly not constant extrapolation as values would be different from the last point for a small interval but close enough as it needs to be C1 smooth everywhere

SouthEndMusic:
I made it now so that both `Constant` and `Extension` extrapolation use `u[end]`, e.g.

```julia
using DataInterpolations
using Plots
using Random

N = 4
Random.seed!(10)
u = rand(N)
t = cumsum(rand(N))
Δt = t[end] - t[1]
t_eval = range(first(t) - Δt, last(t) + Δt, length=500)

A = SmoothedConstantInterpolation(u, t; extrapolation=ExtrapolationType.Extension, d_max=0.2, cache_parameters=true)

plot(t_eval, A.(t_eval), label="interpolation with Extension extrapolation")

scatter!(t, u, label="Data")
```

image

SouthEndMusic:
I don't understand why the derivative tests are failing 🤔

SouthEndMusic:
fixed it 👍

ChrisRackauckas:
I think I'm happy. @sathvikbhagavan ?

docs/src/methods.md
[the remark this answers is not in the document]
sathvikbhagavan: This line can now be removed/edited?

src/interpolation_caches.jl
```diff
@@ -361,6 +366,80 @@ function ConstantInterpolation(
         cache_parameters, assume_linear_t)
 end
 
+"""
+    SmoothedConstantInterpolation(u, t; d_max = Inf, extrapolate = false,
+        cache_parameters = false, assume_linear_t = 1e-2)
+
+It is a method for interpolating constantly with forward fill, with smoothing around the
+value transitions to make the curve continuously differentiable while the integral never
+drifts far from the integral of constant interpolation. `u[end]` is ignored for extrapolation types
```
sathvikbhagavan, line 375: comment about `u[end]` - it can be removed? When is `u[end]` ignored now?

sathvikbhagavan (commented):
A couple of clarifications

SouthEndMusic:
A tangentially related gripe regarding periodic extrapolation: calendars are a mess, making it difficult to periodically extrapolate timeseries in a way that does not get out of sync with them

sathvikbhagavan (approved):
Format and then can be merged!

sathvikbhagavan:
It might be the version. Are you using v1 or v2? CI tries to check in v1

SouthEndMusic:
That did it 🥳 Btw, is there a way to set compat bounds on packages in your base env?

ChrisRackauckas:
Let's update the formatting in a follow up, seems like it takes time to propagate.

ChrisRackauckas (approved):

Issue #373, opened 2025-01-07
1-point ConstantInterpolation gets BoundsError with cache_parameters


visr:
**Describe the bug 🐞**

A `ConstantInterpolation` works fine for single data points as long as extrapolation is enabled. However it cannot be constructed when using `cache_parameters`.

**Expected behavior**

I expect `cache_parameters` to work just like with the setting turned off.

**Minimal Reproducible Example 👇**

```julia
using DataInterpolations
ConstantInterpolation([2], [0.0]; extrapolate = true, cache_parameters = true)
```

**Error & Stacktrace ⚠️**

```
ERROR: BoundsError: attempt to access 0-element Vector{Float64} at index [1]
Stacktrace:
 [1] getindex
   @ .\essentials.jl:13 [inlined]
 [2] first
   @ .\abstractarray.jl:452 [inlined]
 [3] cumulative_integral
   @ a:\.julia\packages\DataInterpolations\C8cNn\src\interpolation_utils.jl:196 [inlined]
 [4] ConstantInterpolation(u::Vector{…}, t::Vector{…}; dir::Symbol, extrapolate::Bool, cache_parameters::Bool, assume_linear_t::Float64)
   @ DataInterpolations a:\.julia\packages\DataInterpolations\C8cNn\src\interpolation_caches.jl:284
```

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```
DataInterpolations v6.6.0
```

  - Output of `versioninfo()`

```
Julia Version 1.10.7
Commit 4976d05258 (2024-11-26 15:57 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 24 × 12th Gen Intel(R) Core(TM) i7-12800HX
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, alderlake)
Threads: 12 default, 0 interactive, 6 GC (on 24 virtual cores)
Environment:
  JULIA_DEPOT_PATH = a:\.julia;
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 12
  JULIA_PKG_PRESERVE_TIERED_INSTALLED = true
  JULIA_PKG_USE_CLI_GIT = true
  JULIA_PROJECT = @.
```

sathvikbhagavan:
I think it is fixed on master.

```julia
[...]

PR #376, opened 2025-01-13
Fix constant interpolation extrapolation

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/373


sathvikbhagavan (commented):
Do we really need to return NaN? If extrapolation is allowed, it is zero anyways

ChrisRackauckas (approved):

DaniGlez:
Hi, would you mind tagging a release with this fix?

PR #387, opened 2025-02-04
ConstantInterpolation Extrapolation

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/378

I started working on this but then I found out this was already fxed by @DaniGlez in https://github.com/SciML/DataInterpolations.jl/pull/379, but the issue wasn't closed. I did some nitpicking taking into account the `dir` of the `ConstantInterpolation` and the edgecase that there is only 1 datapoint.

On a more general note, I see that we use `first` and `last` a lot (also in this PR) which only works properly if `u` is a vector (or at least indexed as such). We also have specialized methods for vector versus matrix `u`. To ensure consistency and avoid code duplication, should we define custom indexing functions based on the type of `u`?

### DIFF
--- a/src/interpolation_methods.jl
+++ b/src/interpolation_methods.jl
@@ -57,7 +57,7 @@ function _extrapolate_left(A::ConstantInterpolation, t)
     if extrapolation_left == ExtrapolationType.None
         throw(LeftExtrapolationError())
     elseif extrapolation_left in (ExtrapolationType.Constant, ExtrapolationType.Linear)
-        first(A.u)
+        dir == :left ? first(A.u) : A.u[min(2, length(A.u))]
     else
         _extrapolate_other(A, t, extrapolation_left)
     end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Add a test for this case?

DaniGlez:
Yeah, it probably makes sense to close the issue, I forgot to use the magic keywords in the PR. Regarding this topic, the only issue I see is that this is consistent with the first / last open interval, but not with evaluation at the edge point, right? Currently, evaluation of a `ConstantInterpolation` at an edge point always returns the edge value regardless of extrapolation and `dir` settings.

```julia-repl
julia> using DataInterpolations
julia> u = [2, 3];
julia> t = [0.0, 1.0];

julia> ConstantInterpolation(u, t; dir=:left)(t)
2-element Vector{Float64}:
 2.0
 3.0

julia> ConstantInterpolation(u, t; dir=:right)(t)
2-element Vector{Float64}:
 2.0
 3.0
```

Incidentally I just realized there seems to be another `eltype` bug with vector inputs:

```julia-repl
julia> ConstantInterpolation(u, t; dir=:right)(t)
2-element Vector{Float64}:
 2.0
 3.0

julia> ConstantInterpolation(u, t; dir=:right).(t)
2-element Vector{Int64}:
 2
 3
```

SouthEndMusic:
@DaniGlez Thanks for pointing out that boundary evaluation behaviour, I made a wrong assumption about it. Then this PR doesn't make sense.

Issue #390, opened 2025-02-06
Error when integrating with `cache_parameters = true`


SouthEndMusic:
**Describe the bug 🐞**

When integrating it seems any interpolation type which was defined with `cache_parameters = true`, in some cases the computation fails with an attempt to evaluate the cached integral vector at index `0`.

**Expected behavior**

Integration without error 😋 

**Minimal Reproducible Example 👇**

```julia
using DataInterpolations

A = LinearInterpolation([1.0, 2.0], [1.0, 2.0]; cache_parameters=true, extrapolation=ExtrapolationType.Extension)
DataInterpolations.integral(A, 0, 1.5)
```

**Error & Stacktrace ⚠️**

```julia
oundsError: attempt to access 1-element Vector{Float64} at index [0]
Stacktrace:
 [1] getindex
   @ .\essentials.jl:13 [inlined]
 [2] integral(A::LinearInterpolation{Vector{…}, Vector{…}, Vector{…}, Vector{…}, Float64, (1,)}, t1::Int64, t2::Float64)
   @ DataInterpolations c:\Users\konin_bt\Repositories\SciML\DataInterpolations.jl\src\integrals.jl:55
```

SouthEndMusic:
So it occurs when `idx2 = 1` which means that no complete interval between data points is in the integration interval

PR #391, opened 2025-02-06
Fix integrating with `cache_parameters = true`

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/390

### DIFF
--- a/test/integral_tests.jl
+++ b/test/integral_tests.jl
@@ -64,7 +64,7 @@ function test_integral(method; args = [], kwargs = [], name::String)
         func, t[1], t[end] + 1.0)
 
     # Test integration with cached parameters
-    if method ∉ [RegularizationSmooth]
+    if method != RegularizationSmooth
         func = method(args...; kwargs..., cache_parameters = true,
             extrapolation = ExtrapolationType.Extension)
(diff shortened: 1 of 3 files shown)


test/integral_tests.jl
```diff
@@ -63,6 +62,15 @@ function test_integral(method; args = [], kwargs = [], name::String)
     @test_throws DataInterpolations.LeftExtrapolationError integral(func, t[1] - 1.0, t[2])
     @test_throws DataInterpolations.RightExtrapolationError integral(
         func, t[1], t[end] + 1.0)
+
+    # Test integration with cached parameters
+    if method != RegularizationSmooth
```
sathvikbhagavan, line 67: Based on your changes, it should also work with RegularizationSmooth?

SouthEndMusic: `RegularizationSmooth` does not support `cache_parameters`

sathvikbhagavan: Oh, can you add it? It can be forwarded to Aitp which is a CubicSpline

sathvikbhagavan (approved):

ChrisRackauckas (approved):

PR #394, opened 2025-02-07
Do not fall back to inplace interpolation method

### LINKED ISSUE #388: Inconsistent output eltype when evaluating on vectors
As of 7.1.0, output eltype is different when evaluating on vectors compared to just broadcasting the scalar path

```julia-repl
julia> using DataInterpolations
julia> u = [2, 3];
julia> t = [0.0, 1.0];
julia> itp = ConstantInterpolation(u, t)
julia> itp(t)
2-element Vector{Float64}:
 2.0
 3.0

julia> itp.(t)
2-element Vector{Int64}:
 2
 3
```

The culprit is in the following lines from https://github.com/SciML/DataInterpolations.jl/blob/master/src/DataInterpolations.jl#L29

```julia
function (interp::AbstractInterpolation)(t::AbstractVector)
    u = get_u(interp.u, t)
    interp(u, t)
end

function get_u(u::AbstractVector, t)
    return similar(t, promote_type(eltype(u), eltype(t)))
end
```

For floats versus ints, a specialization can be defined for `ConstantInterpolation` (whose output is the one that actually preserves the eltype of `u`). But I wonder if the promotion would also generate issues with exotic output types. Would it be possible for `get_u` to use the inferred return type of `interp` itself or would that require usage of arcane Julia internals?

devmotion:
Fixes #388.

### DIFF
--- a/src/DataInterpolations.jl
+++ b/src/DataInterpolations.jl
@@ -25,37 +25,29 @@ include("online.jl")
 include("show.jl")
 
 (interp::AbstractInterpolation)(t::Number) = _interpolate(interp, t)
-
 function (interp::AbstractInterpolation)(t::AbstractVector)
-    u = get_u(interp.u, t)
-    interp(u, t)
-end
-
-function get_u(u::AbstractVector, t)
-    return similar(t, promote_type(eltype(u), eltype(t)))
-end
-
-function get_u(u::AbstractVector{<:AbstractVector}, t)
-    type = promote_type(eltype(eltype(u)), eltype(t))
-    return [zeros(type, length(first(u))) for _ in eachindex(t)]
-end
-
-function get_u(u::AbstractMatrix, t)
-    type = promote_type(eltype(u), eltype(t))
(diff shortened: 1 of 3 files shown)


src/DataInterpolations.jl
```diff
@@ -25,37 +25,29 @@ include("online.jl")
...
+    elseif interp.u isa AbstractArray
+        # Stack interpolated values if `u` was stored in matrix/... form
+        return stack(interp, t)
+    end
 end
 
-function (interp::AbstractInterpolation)(u::AbstractMatrix, t::AbstractVector)
-    @inbounds for i in eachindex(t)
-        u[:, i] = interp(t[i])
+function (interp::AbstractInterpolation)(u::AbstractVector, t::AbstractVector)
```
DaniGlez, line 38: Quick question here, would it be clearer to name this argument something like `out` instead of `u`?

devmotion: Yes, probably. I just kept the existing code here.

ChrisRackauckas (approved):

PR #396, opened 2025-02-10
Remove output dimension type parameter and add `output_dim` to API

devmotion:
The PR removes the type parameter. I tested with my downstream package and tests finally pass again with this change.

The reason is that
- in general, the output dimension cannot be inferred from `u`
- this implies that the constructor of basically all interpolation methods can't be inferred
- I think it's questionable how useful the information is anyway given that output dimension 3 can refer to `u::Vector{<:Vector}` where each element is of length 3 or e.g. a `u::Matrix` with 3 rows

On the master branch:

```julia
julia> using DataInterpolations, Test

julia> @inferred(ConstantInterpolation([1;; 2;; 3], [0.1, 0.2, 0.3]))
ERROR: return type ConstantInterpolation{Matrix{Int64}, Vector{Float64}, Vector{Union{}}, Int64, (1,)} does not match inferred return type ConstantInterpolation{Matrix{Int64}, Vector{Float64}, Vector{T}, Int64} where T
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[13]:1

julia> @inferred(ConstantInterpolation(["A", "B", "C"], [0.1, 0.2, 0.3]))
ERROR: return type ConstantInterpolation{Vector{String}, Vector{Float64}, Nothing, String, (1,)} does not match inferred return type ConstantInterpolation{Vector{String}, Vector{Float64}, Nothing, String}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[14]:1

julia> @inferred(ConstantInterpolation([[1], [2], [3]], [0.1, 0.2, 0.3]))
ERROR: return type ConstantInterpolation{Vector{Vector{Int64}}, Vector{Float64}, Nothing, Vector{Int64}, (1,)} does not match inferred return type ConstantInterpolation{Vector{Vector{Int64}}, Vector{Float64}, Nothing, Vector{Int64}}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[15]:1
```

This PR:
```julia
julia> using DataInterpolations, Test

julia> @inferred(ConstantInterpolation([1;; 2;; 3], [0.1, 0.2, 0.3]))
ConstantInterpolation with 3 points, in left direction
┌──────┬─────┐
│ time │  u1 │
├──────┼─────┤
│  0.1 │ 1.0 │
│  0.2 │ 2.0 │
│  0.3 │ 3.0 │
└──────┴─────┘


julia> @inferred(ConstantInterpolation(["A", "B", "C"], [0.1, 0.2, 0.3]))
ConstantInterpolation with 3 points, in left direction
┌──────┬───┐
│ time │ u │
├──────┼───┤
│  0.1 │ A │
│  0.2 │ B │
│  0.3 │ C │
└──────┴───┘


julia> @inferred(ConstantInterpolation([[1], [2], [3]], [0.1, 0.2, 0.3]))
ConstantInterpolation with 3 points, in left direction
┌──────┬─────┐
│ time │  u1 │
├──────┼─────┤
│  0.1 │ 1.0 │
│  0.2 │ 2.0 │
│  0.3 │ 3.0 │
└──────┴─────┘
```

---

Edit: I added an exported `output_dim` function as discussed in https://github.com/SciML/DataInterpolations.jl/pull/396#issuecomment-2648396729. This could be used by SparseConnectivityTracer if desired/useful which currently seems to be the only downstream user of the type parameter but only makes us of it in a non-static way.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataInterpolations"
 uuid = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
-version = "7.2.0"
+version = "8.0.0"
 
 [deps]
 EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,4 @@
 [deps]
-DataInterpolations = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 ModelingToolkitStandardLibrary = "16a59e39-deab-5bd0-87e4-056b12336739"
@@ -11,7 +10,6 @@ StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 
 [compat]
-DataInterpolations = "6, 7.0"
 Documenter = "1"
 ModelingToolkit = "9"
 ModelingToolkitStandardLibrary = "2"
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -19,4 +19,5 @@ QuinticHermiteSpline
 
 ```@docs
 DataInterpolations.looks_linear
+DataInterpolations.output_dim
 ```
--- a/src/DataInterpolations.jl
+++ b/src/DataInterpolations.jl
@@ -98,11 +98,23 @@ function Base.showerror(io::IO, ::ExtrapolationNotImplementedError)
     print(io, EXTRAPOLATION_NOT_IMPLEMENTED_ERROR)
 end
 
+"""
+    output_dim(x::AbstractInterpolation)
+
+Return the number of dimension `ndims(x(t))` of interpolation `x` evaluated at a single value `t`
+if `x(t) isa AbstractArray`, or 0 otherwise.
+"""
+output_dim(x::AbstractInterpolation) = _output_dim(x.u)
+_output_dim(::AbstractVector) = 0 # each value is a scalar
+_output_dim(::AbstractVector{<:AbstractArray{<:Any, N}}) where {N} = N # each value is an array but values are not stacked
+_output_dim(::AbstractArray{<:Any, N}) where {N} = N - 1 # each value is an array but multiple values are stacked
+
 export LinearInterpolation, QuadraticInterpolation, LagrangeInterpolation,
        AkimaInterpolation, ConstantInterpolation, QuadraticSpline, CubicSpline,
        BSplineInterpolation, BSplineApprox, CubicHermiteSpline, PCHIPInterpolation,
        QuinticHermiteSpline, LinearInterpolationIntInv, ConstantInterpolationIntInv,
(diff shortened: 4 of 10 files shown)


SouthEndMusic:
@ChrisRackauckas isn't this another argument in favor of the `SimpleInterpolations.jl` you suggested?

sathvikbhagavan:
I had added it as a feature request (https://github.com/SciML/DataInterpolations.jl/issues/328) but if it breaks inference, I guess its not worth it.

devmotion:
It's used https://github.com/adrhill/SparseConnectivityTracer.jl/blob/622cfb729832d127f2d0f9f94110391b4d439b6b/ext/SparseConnectivityTracerDataInterpolationsExt.jl#L54 and https://github.com/adrhill/SparseConnectivityTracer.jl/blob/622cfb729832d127f2d0f9f94110391b4d439b6b/ext/SparseConnectivityTracerDataInterpolationsExt.jl#L64 but these cases do not need a static type information but can just extract the dimension (IMO still questionable how useful in general given the different types of `u`).

ChrisRackauckas:
Why is it not inferable? This seems trivial to infer?

devmotion:
No, it can only be inferred from static arrays.

ChrisRackauckas:
Wait it's not a dimension size computable from `ndims(u)`? I guess I haven't looked at it in awhile but it should be as simple as that.

ChrisRackauckas:
Oh, that can't be type information. Wait why would DI even want or need that? That seems silly.

devmotion:
🤷

SouthEndMusic:
From the original issue by @adrhill:

> As a developer writing a package extension for DataInterpolations.jl (https://github.com/adrhill/SparseConnectivityTracer.jl/pull/178), it would be useful if all AbstractInterpolations indicated their output dimensionality, e.g.

So it might be useful to use `get_output_dim` in the DI extension directly, so that DI doesn't have to know explicitly how different types of `u` are sliced by `DataInterpolations.jl`? Or should they just infer it from an evaluation?

adrhill:
> Wait it's not a dimension size computable from `ndims(u)`? I guess I haven't looked at it in awhile but it should be as simple as that.

I thought this as well. At that was my intention in opening #328. I have no problem with a revert of whatever `N` is currently.

ChrisRackauckas:
Let's fix it to `ndims`. This seems to just be a mistake.

devmotion:
So, you'd want
```julia
output_dims(::AbstractVector) = 0 # each value is a scalar
output_dims(::AbstractVector{<:AbstractArray{<:Any,N}}) where {N} = N # each value is an array but values are not stacked
output_dims(::AbstractArray{<:Any,N}) where {N} = N - 1 # each value is an array but multiple values are stacked
```

SouthEndMusic:
@ChrisRackauckas friendly bump to merge this and create a new release 🙂

ChrisRackauckas:
Wasn't the final say here to just make it match `ndims` and the current code doesn't do that?

SouthEndMusic:
So to have the output dimensionality as the type parameter instead of having the function `output_dims`? I'd say both are fine, but the first one is a bit nicer.

ChrisRackauckas (approved):

Issue #399, opened 2025-03-06
`PCHIPInterpolation` method does not produce the same results as other interpolation libraries


egavazzi:
**Describe the bug 🐞**

The `PCHIPInterpolation` method from `DataInterpolations.jl` produces results that are slightly different than what the `pchip` methods from MATLAB, SciPy and PCHIPInterpolation.jl  produce.

**Expected behavior**

The results of `PCHIPInterpolation` should be identical to the other libraries.

**Minimal Reproducible Example 👇**

Notes: 
- Calling Python and Matlab requires some extra setup. But comparing with `PCHIPInterpolation.jl` should be easy.
- I reused the numbers from my real world example where I noticed the difference, that's why they are a bit "funny"

```julia
import DataInterpolations
using PythonCall
using MATLAB
import PCHIPInterpolation

x = [-1.529205842868462, -1.401759744308854, -1.298455005566862,
    -1.229836683674959, -1.107192298263224, -0.930884014999626,
    -0.739165582563789, -0.534726081040409, -0.361197419278502,
    -0.210753800802626, -0.023659016727907, 0.148387783842545,
    0.195027357403972, 0.256843155275395, 0.271309779815279,
    0.284961440170103, 0.292280091257079, 0.313103310290649,
    0.339325032538764, 0.360990712227029, 0.383915274219258,
    0.406241931082478, 0.426693754073074, 0.438007445340180,
    0.442409564564855, 0.449441512598424, 0.461625748593174,
    0.475071834892396, 0.479153568268782, 0.698970004336019,
    1.000000000000000, 1.301029995663981, 1.477121254719663,
    1.698970004336019]

y = [-16.489672666179871, -16.255790604873923, -16.087003295315228,
    -15.996264950740080, -15.953203452818123, -15.989191402537488,
    -16.041042208514060, -16.116569595034353, -16.168214419287263,
    -16.193450575738691, -16.195569244896330, -16.150452456808626,
    -16.107709885163285, -16.025430820560523, -15.961620468316118,
    -15.840488804947917, -15.803011086935593, -15.948567729023981,
    -15.533186544559602, -15.883157959631777, -15.476897837253116,
    -15.795237176295828, -15.528433895057361, -15.719342640373487,
    -15.765657146956446, -15.706366402956451, -15.622036628514593,
    -15.639548788652569, -15.655341886713167, -15.511449283499555,
    -15.414539270491499, -15.468521082957746, -15.552841968657781,
[...]

PR #400, opened 2025-03-07
Fix PCHIP du computation

SouthEndMusic:
Fixes https://github.com/SciML/DataInterpolations.jl/issues/399

Difference between `DataInteprolations.jl` and `PCHIPInterpolation.jl` for the data in the issue:

image


sathvikbhagavan (approved):

egavazzi:
Wow, that was fast!! Thank you!
Small differences on the order of 1e-15 is what I also observed between the SciPy/Matlab/PCHIPInterpolations.jl implementations.

Issue #401, opened 2025-03-07
Structural non-allocating array-like interpolation


SouthEndMusic:
I have the following 🍇  about the current implementation:
- Array-valued interpolations are allocating;
- Different methods for scalar and array interpolation per interpolation type yields code duplication;
- Support for array-valued interpolation is inconsistent.

As far as I'm aware, for every interpolation type the array-valued interpolation can be defined by a per-index scalar interpolation.

I propose:
- Array Interpolation objects get a pre-allocated value, derivative and integral on which computations are done in-place;
- We define a function `get_u(A.u, i)` which gets the appropriate view of a slice of `u`.

Related issues:
- https://github.com/SciML/DataInterpolations.jl/issues/118
- https://github.com/SciML/DataInterpolations.jl/issues/361

SouthEndMusic:
How so?

I didn't say that to inform the implementation btw, more to have a consistent way of defining what array-valued interpolation means.

sathvikbhagavan:
> I didn't say that to inform the implementation btw, more to have a consistent way of defining what array-valued interpolation means.

Oh, ok. I was thinking having an interpolation object for every time series for higher order arrays might be expensive in both memory and time. Hence, it might be better to have separate dispatches for array valued where everything is vectorized.

SouthEndMusic:
That's something that I wondered about: is it better to use different code for scalar and array cases, and why is that?

ChrisRackauckas:
You cannot write in-place code for scalars since they are immutable. Scalar code can be non-allocating while having caches for arrays cannot be. So if you want both to be non-allocating, you need different code for both.

> Array-valued interpolations are allocating;

The in-place form is not supposed to be?

SouthEndMusic:
I set up `NDInterpolations.jl` in a way that satisfies my points here. There's a single function per method for scalar and vector valued interpolation, and in place vector valued interpolation is supported.

PR #404, opened 2025-03-16
Smooth arclength interpolation

SouthEndMusic:
Over the past few weeks, in my hobby math and coding time I have been fixated on a question that arose [here](https://github.com/JuliaHealth/BloodFlowTrixi.jl/issues/20), namely: _What is a good way to interpolate through a set of points in_ $\mathbb{R}^n, n \ge 2$ _, which is_ $C^1$ _smooth __and__ parameterized by arc length?_

I understood that a common approach here is to make an interpolation that is not by arc-length and then approximate the arc-length parameterization using numerical methods. I however figured there should be a way to do this without such heavy lifting. I tried several algebraic approaches which didn't really get me anywhere, so then I started looking for a more geometrically informed approach.

A quick Google search reveals [this pathological method](https://math.stackexchange.com/a/4458846), which suffers from not being at all flexible, the effect of points being non-local, and having arbitrarily large detours if the points are distributed in an unfortunate way. 

The core idea I had is to again start with a non arc-length parameterized curve, but then to approximate it with segments that are trivially parameterizable by arc-length, namely line- and circle-segments. I explain the full construction of the method in [a notebook that I checked into this branch](https://github.com/SouthEndMusic/DataInterpolations.jl/blob/smooth_arclength_interpolation/arclength_interpolation.ipynb).

I am not sure this method is suitable for `DataInterpolations.jl`, but it can make good use of the internals, so maybe it should be a separate package depending on `DataInterpolations.jl`.

I now use `QuadraticSpline` as the default 'shape-defining' curve, because it is a method that is $C^1$ smooth and supports vector valued interpolation. However, for this method also the effect of data points is non-local, which was my motivation for opening https://github.com/SciML/DataInterpolations.jl/issues/401.

I am btw not completely sure what to put in the docstring of `struct SmoothArcLengthInterpolation`, because there are several different methods which are different entrypoints in the construction of the interpolation.

image


yolhan83:
Wow, you really went through that 👍 for usefulness, for now I only really got the case of advection simulation on the curve where you really need that, I don't know if there are others.

ChrisRackauckas (dismissed):
Requires a squash merge to make sure the memory of the notebook isn't in the core repo.

ChrisRackauckas:
Just realized, this is missing tests.

SouthEndMusic:
Yep, I'm on it

ChrisRackauckas (approved):

PR #407, opened 2025-03-20
Fix type instabilities from extrapolation

SouthEndMusic:
I made these changes because I saw some type instabilities when evaluating `LinearInterpolationIntInv`. My hypothesis is that type inference failed because of an infinite recursion (that doesn't happen at runtime). At any rate, the extrapolation functionality now uses a bit more knowledge from those code paths for evaluation too.


ChrisRackauckas:
Add some inference tests?

ChrisRackauckas (approved):

PR #413, opened 2025-03-24
Add `output_size`

visr:
I wanted to update SparseConnectivityTracer to DataInterpolations v8 in https://github.com/adrhill/SparseConnectivityTracer.jl/pull/234 but got a bit confused.

https://github.com/SciML/DataInterpolations.jl/pull/396 removes the output dimension type parameter and the old `get_output_dim`. Even though it talks about dimensions it was really the `size` of the output.

SparseConnectivityTracer was one of the reasons `output_dim` was added to the API, but this actually was the `ndims` of the output (perhaps `output_ndims` would be a better name). But that is not what SparseConnectivityTracer needs. Hence I am proposing to add `output_size`, such that this code doesn't have to live in SparseConnectivityTracer.

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -20,4 +20,5 @@ QuinticHermiteSpline
 ```@docs
 DataInterpolations.looks_linear
 DataInterpolations.output_dim
+DataInterpolations.output_size
 ```
--- a/src/DataInterpolations.jl
+++ b/src/DataInterpolations.jl
@@ -101,20 +101,29 @@ end
 """
     output_dim(x::AbstractInterpolation)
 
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
`length` should be the number of points and `size` the size of an output?

visr:
Sounds reasonable, but not sure what your comment means for this PR.

visr:
Converted to draft until we are sure this is what SparseConnectivityTracer really needs, see https://github.com/adrhill/SparseConnectivityTracer.jl/pull/234#discussion_r2011996798.

ChrisRackauckas:
😅  Just make a new PR?

visr:
No need, despite the red closed icon, it is on master as intended.

PR #418, opened 2025-04-08
remove unnecessary recipes

oscardssmith:
NFC. According to Chris, there used to be an issue defining recipes for callable objects, but that issue hasn't existed for years at this point.


sathvikbhagavan (approved):
Yeah, I have not seen people use it.

sathvikbhagavan:
Can you also remove these from the documentation and README as well?

PR #421, opened 2025-04-09
Ensure that the interpolation domain is sorted

SebastianM-C:
## Checklist

## Additional context

A frequent user error is forgetting the correct argument error. This results in silently wrong results or extrapolation errors. I think that we can surface this error earlier to avoid further confusion. Let me know what you think about this.

### DIFF
--- a/src/interpolation_utils.jl
+++ b/src/interpolation_utils.jl
@@ -91,14 +91,15 @@ function quadratic_spline_params(t::AbstractVector, sc::AbstractVector)
 end
 
 # helper function for data manipulation
-function munge_data(u::AbstractVector, t::AbstractVector)
+function munge_data(u::AbstractVector, t::AbstractVector;
+        check_sorted = t, sorted_arg_name = "second")
     Tu = nonmissingtype(eltype(u))
     Tt = nonmissingtype(eltype(t))
 
     if Tu === eltype(u) && Tt === eltype(t)
-        if !(issorted(t) || issorted(t, rev = true))
+        if !(issorted(check_sorted) || issorted(check_sorted, rev = true))
             # there is likely an user error
-            msg = "The second argument, which is used for the interpolation domain, is not sorted."
+            msg = "The $sorted_arg_name argument, which is used for the interpolation domain, is not sorted."
             if (issorted(u) || issorted(u, rev = true))
                 msg *= "\nIt looks like the arguments were inversed, make sure you used the arguments in the correct order."
             end
--- a/src/interpolation_utils.jl
+++ b/src/interpolation_utils.jl
@@ -92,16 +92,16 @@ end
 
 # helper function for data manipulation
 function munge_data(u::AbstractVector, t::AbstractVector;
-        check_sorted = t, sorted_arg_name = "second")
+        check_sorted = t, sorted_arg_name = ("second", "t"))
     Tu = nonmissingtype(eltype(u))
     Tt = nonmissingtype(eltype(t))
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas (dismissed):

SebastianM-C:
Sorry, got the function wrong :sweat_smile: 
Should be fine now, I hope. I also noticed that we didn't check that `length(t) == length(u)` if there were no missing values, so I moved that up.

SebastianM-C:
hmm, It looks like `RegularizationSmooth` might need special treatment since it has 3 argument and the 3rd one has to be sorted, not the second one.

src/interpolation_utils.jl
```diff
@@ -91,14 +91,26 @@ function quadratic_spline_params(t::AbstractVector, sc::AbstractVector)
...
+        check_sorted = t, sorted_arg_name = "second")
     Tu = nonmissingtype(eltype(u))
     Tt = nonmissingtype(eltype(t))
+
     if Tu === eltype(u) && Tt === eltype(t)
+        if !(issorted(check_sorted) || issorted(check_sorted, rev = true))
+            # there is likely an user error
+            msg = "The $sorted_arg_name argument, which is used for the interpolation domain, is not sorted."
+            if (issorted(u) || issorted(u, rev = true))
+                msg *= "\nIt looks like the arguments were inversed, make sure you used the arguments in the correct order."
```
SouthEndMusic, line 104: I think it would be more clear if you specifically refer to `u` and `t` in addition to referring to the ordering of the args.

src/interpolation_utils.jl
```diff
@@ -91,14 +91,26 @@ function quadratic_spline_params(t::AbstractVector, sc::AbstractVector)
...
 
 # helper function for data manipulation
-function munge_data(u::AbstractVector, t::AbstractVector)
+function munge_data(u::AbstractVector, t::AbstractVector;
+        check_sorted = t, sorted_arg_name = "second")
     Tu = nonmissingtype(eltype(u))
     Tt = nonmissingtype(eltype(t))
+
     if Tu === eltype(u) && Tt === eltype(t)
+        if !(issorted(check_sorted) || issorted(check_sorted, rev = true))
```
SouthEndMusic, line 100: Should we support reverse ordering? It wouldn't surprise me if that violates an implicit assumption in certain methods.

SouthEndMusic (approved):
Nice, I remember being somewhat surprised that we didn't validate for the ordering of `t` before. I left a few comments.

src/interpolation_utils.jl
SebastianM-C: Let me see if there's actually any test that errors if I remove reverse ordering. Some methods explicitly mention that they need monotonically increasing values.

SebastianM-C: [842f606](https://github.com/SciML/DataInterpolations.jl/pull/421/commits/842f6061833a673479afccdb321d207d004bdf21) changes to just monotonically increasing, nothing seems to be using reverse ordering and as you mention it likely is wrong. Is that would be needed in the future, we can add another flag to the function for that.

src/interpolation_utils.jl
SebastianM-C: Is [b419397](https://github.com/SciML/DataInterpolations.jl/pull/421/commits/b419397e8524b1fcbdeec28936d6462cbafe7b60) okay?

SouthEndMusic (approved):

PR #428, opened 2025-05-02
use isequal instead of == in `safe_diff`

vyudu:
This would make interpolation on symbolic arrays work, e.g. 
```julia
@variables x[1:101]
t = 0.:0.01:1
LinearInterpolation(x, t)
```
would work as expected (would throw errors before).

### DIFF
--- a/src/parameter_caches.jl
+++ b/src/parameter_caches.jl
@@ -15,7 +15,7 @@ end
 
 # Prevent e.g. Inf - Inf = NaN
 function safe_diff(b, a::T) where {T}
-    b == a ? zero(T) : b - a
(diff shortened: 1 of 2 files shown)


vyudu:
@ChrisRackauckas this would make things nicer for CasADi symbolic objects because it doesn’t have built in interpolation, so being able to just use DataInterpolations machinery would be nice

ChrisRackauckas (approved):

vyudu:
Marked out the ones that work and don't in the test, should I document this soemwhere?

ChrisRackauckas:
Open an issue to follow up on those.

ChrisRackauckas (approved):

Issue #419, opened 2025-04-08
Bsplines seem slightly broken


oscardssmith:
```
julia> using DataInterpolations, Plots
julia> using DataInterpolations: derivative
julia> u = [2.0, 1.0, 5.0, 4.0, 5.0, 4.0]
julia> t = [0.0, 2.0, 3.5, 4.0, 5.0, 6.5]
julia> bspline = BSplineInterpolation(u, t, 0, :Uniform, :Uniform; extrapolation_left=ExtrapolationType.Extension,  extrapolation_right=ExtrapolationType.Extension)
julia> plot(bspline)
```
![Image](https://github.com/user-attachments/assets/f2111ad1-a81d-45eb-a8fd-3f3265e42c14)
Why isn't this matching ConstantInterpolation (e.g. having discontinuities at the data points)

For extra weirdness:
```
bspline = BSplineInterpolation(u, t, 2, :Uniform, :Uniform; extrapolation_left=ExtrapolationType.Extension,  extrapolation_right=ExtrapolationType.Extension)
plot(t->derivative(bspline, t))
```
![Image](https://github.com/user-attachments/assets/854e2772-4f5f-48d4-9938-65ca12db45a8)

ChrisRackauckas:
@sathvikbhagavan I think you were looking at this last?

sathvikbhagavan:
I will look into it

SouthEndMusic:
Regarding the first plot: the discontinuities are at the knots. These knots are computed in the constructor, although I've never delved deeply into what actually happens there. In the BSpline interpolation I'm working on in `NDInterpolations.jl`, the knots will coincide with `t`.

oscardssmith:
yeah, that feels like the right answer

SouthEndMusic:
Now that I delved a bit deeper into B-Splines again for `NDInteprolations.jl`, I understand this a bit more. `DataInterpolations.jl` generally assumes that the input data consists of points `(t,u)`. However, for B-Splines this is a bit different; B-Splines have knots and control points. Control points are not associated with points in the domain, but with basis functions which have support on a small number of intervals between knots.

Generally the number of distinct knots is not equal to the number of control points, which breaks the aforementioned assumption. The `BSplineInterpolation` somehow ties things together, in a way which is not explained in the doc(string)s but there is a reference: https://pages.mtu.edu/%7Eshene/COURSES/cs3621/NOTES/spline/B-spline/bspline-curve.html.

[...]

PR #440, opened 2025-07-09
support ForwardDiff@1.0

oscardssmith:
This adds support for ForwardDiff@1.0, but removes tests for derivatives of Bplines, since their behavior (as pointed out in https://github.com/SciML/DataInterpolations.jl/issues/419 is already fairly buggy).

I think this is the right balance to strike, as it turns out the behavior here was always fairly wonky.

### DIFF
--- a/test/derivative_tests.jl
+++ b/test/derivative_tests.jl
@@ -6,7 +6,7 @@ using Symbolics
 using StableRNGs
 using RegularizationTools
 using Optim
-using ForwardDiff
+import ForwardDiff
 using LinearAlgebra
 
 function test_derivatives(method; args = [], kwargs = [], name::String)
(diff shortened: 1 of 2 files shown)


test/derivative_tests.jl
```diff
@@ -35,11 +35,8 @@ function test_derivatives(method; args = [], kwargs = [], name::String)
 
         # Interpolation transition points
         for _t in t[2:(end - 1)]
-            if func isa Union{BSplineInterpolation, BSplineApprox,
-                CubicHermiteSpline}
-                fdiff = forward_fdm(5, 1; geom = true)(func, _t)
-                fdiff2 = forward_fdm(5, 1; geom = true)(t -> derivative(func, t), _t)
-            elseif func isa SmoothedConstantInterpolation
+            if func isa Union{SmoothedConstantInterpolation, BSplineInterpolation, BSplineApprox}
+                # TODO fix interpolations
```
ChrisRackauckas, line 39: use `@test_broken` instead?

oscardssmith: fair.

oscardssmith: the tricky part is that some of these work (typically if the derivative is the same before vs after). As such, `@test_broken` also fails.

PR #441, opened 2025-07-10
Add smoothed linear interpolation

SouthEndMusic:
## Checklist

## Additional context

I first described this method [here](https://southendmusic.github.io/CachedInterpolations.jl/stable/construction/construction_smoothed_linear_interpolation/). I'm not sure whether I should put all of that in the DataInterpolations docs.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -20,6 +20,7 @@ corresponding to `(u,t)` pairs.
 
   - `SmoothedConstantInterpolation(u,t)` - An integral preserving continuously differentiable approximation of constant interpolation.
   - `LinearInterpolation(u,t)` - A linear interpolation.
+  - `SmoothedLinearInterpolation(u,t;  λ = 0.25)` - A continuously differentiable linear interpolation with an interval around each data point replaced by spline section.
   - `QuadraticInterpolation(u,t)` - A quadratic interpolation.
   - `LagrangeInterpolation(u,t,n)` - A Lagrange interpolation of order `n`.
   - `QuadraticSpline(u,t)` - A quadratic spline interpolation.
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -2,6 +2,7 @@
 
 ```@docs
 LinearInterpolation
+SmoothedLinearInterpolation
 QuadraticInterpolation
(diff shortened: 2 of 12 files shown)


SouthEndMusic:
A question related to SparseConnectivityTracer: The DataInterpolations extension there handles each interpolation type individually, so new interpolation types do not work out of the box with SparseConnectivityTracer. Specifically, `SparseConnectivityTracer` needs to know whether the first and second derivative of an interpolation type are identically zero. Would it be ok with you @sathvikbhagavan @ChrisRackauckas if this information lives in DataInterpolations instead of in the SparseConnectivityTracer extension for DataInterpolations?

sathvikbhagavan:
> The alternative would be to move the extension from SCT to this repo.

I am fine moving the extension here if that helps.

adrhill:
Tagged SCT `v1.0.0` without the extension.

sathvikbhagavan (dismissed):
LGTM! Just needs a run with the formatter.

adrhill:
> @adrhill do you know why some of the SCT tests for the new interpolation type are failing?

The tests only fail on local Hessian sparsity patterns using SCT's `Dual` numbers.
Up until now, the test suite was able to make the assumption that for the [given inputs `t`, `u`](https://github.com/SciML/DataInterpolations.jl/blob/c1815967d6be25a1ca30f9933e7c722cd7fe81d0/test/sparseconnectivitytracer_tests.jl#L17-L25) etc., interpolations have **either** zero or non-zero second-order derivatives everywhere. That's how the reference Hessian is determined:

https://github.com/SciML/DataInterpolations.jl/blob/c1815967d6be25a1ca30f9933e7c722cd7fe81d0/test/sparseconnectivitytracer_tests.jl#L134

`SmoothedLinearInterpolation` breaks that pattern by having 2nd-order derivatives that are non-zero around datapoints but zero otherwise. So you are going to have to update the tests, or at least `Href`.

adrhill:
> The formatter (I use `JuliaFormatter 2.1.2` locally) doesn't converge on the SCT extension

I'm guessing the extension still had SCT's Runic formatting applied. 🤷‍♂️

PR #457, opened 2025-08-10
Convert `z` in `CubicSpline` for performance

cgarling:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context

This hopes to address the performance issues of multi-valued interpolation discussed in #431 by using dependent variable arrays with elements of `StaticArrays.SArray` to remove allocations and improve performance. On `master` this still allocates when evaluating the interpolants, hurting performance.

For the example of `CubicSpline`, I isolated the allocations to come from the `z` array. By converting the elements of `z` to match `eltype(u)`, we are able to remove allocations when evaluating the interpolant and greatly improve the performance. Example usage and performance comparison is given below. 

All tests pass for me locally on Julia 1.11.6. If this seems like a reasonable solution, I can look into modifying the other relevant methods as well.

```julia
using StaticArrays: SVector, SArray
using DataInterpolations: CubicSpline
using BenchmarkTools: @benchmark
using  Test: @test

x = 1:1_000
a,b,c = [rand(length(x)) for i in 1:3]
y = [SVector{3, Float64}(a[i], b[i], c[i]) for i in eachindex(a,b,c)]

cubic = CubicSpline(y, x)
@benchmark $cubic($2.5)
# On master, ↑ is 133.750 ns with Memory estimate: 640 bytes, allocs estimate: 16.
# On PR: 21.717 ns, 0 allocations

y2 = [SArray{Tuple{1,2,3}, Float64, 3, 6}(rand(2,3)) for i in eachindex(a,b,c)]
cubic2 = CubicSpline(y2, x)
@benchmark $cubic2($2.5)
# On master, ↑ is 248.704 ns with Memory estimate: 1.00 KiB, allocs estimate: 16.
# On PR: 25.965 ns ns, 0 allocations

# Vector of vectors
y3 = [[a[i], b[i], c[i]] for i in eachindex(a,b,c)]
cubic3 = CubicSpline(y3, x)
@benchmark $cubic3($2.5)
# On master, ↑ is 243.811 ns with Memory estimate: 1.17 KiB, allocs estimate: 30.
[...]

PR #466, opened 2025-09-03
Test allocations

cgarling:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
- [ ] Any new documentation only uses public API
  
Following #457, this PR adds more tests for allocation-free interpolation of `Vector{SVector}` inputs. To my knowledge, all interpolators that work with vector-valued inputs now have these tests. That list is
 - CubicSpline
 - QuadraticInterpolation
 - SmoothedConstantInterpolation
 - ConstantInterpolation
 - QuadraticSpline
 - LinearInterpolation
 - CubicHermiteSpline
 - QuinticHermiteSpline

Two minor code changes are included to support this work. 

- In `spline_coefficients!` it is required to assert the return type of `findfirst(x -> x > u, k)::Int` to avoid a dynamic dispatch to `-` as `findfirst` can return an `Int` or `Nothing`. The dynamic dispatch causes the allocation check via AllocCheck.jl to fail. As bounds checks on the inputs are performed during the interpolation step, I think it is safe to assert this return type.
- The last branch in `smoothed_constant_interpolation_parameters` does not work on array-valued inputs as `one(::AbstractArray)` is not defined. I think type stability can be maintained by replacing `zero(one(eltype(u)) / 2)` with `zero(first(u) / 2)`, but this is maybe not optimal if `u` has elements of mixed type. I suppose it could be `zero(u[idx] / 2)`, other suggestions welcome.

Also some of CubicSpline `@inferred` tests that are labeled broken are now passing for me locally (Julia 1.11) so I changed those. If they fail on CI we can troubleshoot further.

### DIFF
--- a/src/interpolation_utils.jl
+++ b/src/interpolation_utils.jl
@@ -38,7 +38,7 @@ function spline_coefficients!(N, d, k, u::Number)
         N[end] = one(u)
         return length(N):length(N)
     else
-        i = findfirst(x -> x > u, k) - 1
+        i = findfirst(x -> x > u, k)::Int - 1
         N[i] = one(u)
         for deg in 1:d
             N[i - deg] = (k[i + 1] - u) / (k[i + 1] - k[i - deg + 1]) * N[i - deg + 1]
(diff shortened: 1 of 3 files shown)


cgarling:
Three CubicSpline tests there were marked as broken on Julia < 1.11 are now passing on LTS ( see https://github.com/SciML/DataInterpolations.jl/actions/runs/17439841023/job/49519664698?pr=466) so I've updated those as well.

ChrisRackauckas (approved):

PR #470, opened 2025-09-15
Fix + test Integrating `SmoothedConstantInterpolation` without smoothing

SouthEndMusic:


ChrisRackauckas (approved):

ChrisRackauckas:
Don't know why you wouldn't just use ConstantInterpolation there 😅 but it seems fine.

PR #474, opened 2025-09-30
bump version

oscardssmith:
for PrettyTables


ChrisRackauckas:
Are you sure it didn't require any change?

PR #506, opened 2026-02-19
refactor: migrate Curvefit from Optim.jl to CurveFit.jl

sathvikbhagavan:
fixes: #505


sathvikbhagavan:
Hi @ChrisRackauckas, how do we pass bounds using CurveFit? CurveFitProblems https://github.com/SciML/CurveFit.jl/blob/master/src/common_interface.jl#L49 does not take any bounds?

ChrisRackauckas:
Let's rebase and merge the breaking.

PR #511, opened 2026-04-06
refactor: improve BSplines with simpler parameterization

sathvikbhagavan:
This PR simplifies BSplines. There is no need to have intermediate parameterization p ∈ [0,1] and a piecewise-linear mapping t → p, which is designed for 2D/3D parametric curves, not function interpolation.


ChrisRackauckas (dismissed):

ChrisRackauckas:
Register the new major?

PR #467, opened 2025-09-07
Add missing matrix support for AkimaInterpolation and QuadraticSpline

ChrisRackauckas-Claude:
## Summary

Fixes #206 by implementing missing matrix dispatch methods for `AkimaInterpolation` and `QuadraticSpline`.

Previously, only `LinearInterpolation`, `QuadraticInterpolation`, `LagrangeInterpolation`, and `ConstantInterpolation` supported matrices. This PR adds matrix support to the remaining methods.

## Changes Made

- **AkimaInterpolation**: Added `AbstractMatrix` constructor and `_interpolate` method
- **QuadraticSpline**: Added `AbstractMatrix` constructor and `_interpolate` method

## Matrix Interface

The matrix interface allows interpolating multiple functions simultaneously, where each row of the matrix represents a different function evaluated at the same time points.

Example usage:
```julia
t = 0.0:0.1:1.0
u_matrix = [sin.(t) cos.(t)]'  # 2×11 matrix (2 functions, 11 time points)

# Now works for all methods:
itp1 = AkimaInterpolation(u_matrix, t)
itp2 = QuadraticSpline(u_matrix, t)

result = itp1(0.5)  # Returns [sin(0.5), cos(0.5)] approximately
```

## Test Results

All methods now have consistent matrix support:
- ✅ LinearInterpolation, QuadraticInterpolation, LagrangeInterpolation, ConstantInterpolation (already worked)
- ✅ AkimaInterpolation, QuadraticSpline (newly fixed) 
- ✅ CubicSpline (already worked)

## Notes

- BSplineInterpolation and BSplineApprox still have constructor issues, but these appear to be separate bugs not related to matrix support
- Vector of Vectors support for AkimaInterpolation remains a separate issue due to the `abs` function not working with vector elements

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #521, opened 2026-05-06
Add matrix support for AkimaInterpolation

### LINKED ISSUE #468: Add vector output support for AkimaInterpolation
**Is your feature request related to a problem? Please describe.**
Currently AkimaInterpolation only supports output of a single value so it cannot output a vector of values.  This support is available for LinearInterpolation, CubicSpline, etc.

A clear and concise description of what the problem is. Ex. I’m always frustrated when […]
AkimaInterpolation only supports a single vector for interpolatin unlike some of the other methods implemented.

**Describe the solution you’d like**

The following should work as it does for other methods:
u = vcat(2.0collect(1:10)', 3.0collect(1:10)')
t = 1.0collect(1:10)
A = AkimaInterpolation(u, t)

**Additional context**

It might be nice if the docs indicate which methods support vector output and which do not.

mrarat:
## Summary

This PR adds `AbstractMatrix` input support for `AkimaInterpolation`.

Previously, `AkimaInterpolation(u, t)` only accepted vector-like `u`, while the documented interface supports stacked array data where the last dimension corresponds to time. With this change, matrix-valued data can be passed as:

```julia
u::AbstractMatrix  # rows are output components, columns correspond to t
t::AbstractVector
A = AkimaInterpolation(u, t)
```
For scalar evaluation times, A(t_eval) now returns a vector containing the interpolated value for each row of u.

## Changes
- Added an AkimaInterpolation(u::AbstractMatrix, t; ...) constructor.
- Added matrix-valued Akima interpolation evaluation.
- Added matrix-valued Akima derivative support for in-domain evaluation.
- Added matrix-valued Akima integral support for in-domain integration.
- Added tests comparing matrix Akima behavior against independent row-wise scalar Akima interpolations.

## Known limitation

This PR intentionally does not fix matrix-valued extrapolation.

The current generic extrapolation code in DataInterpolations uses first(A.u) and last(A.u) to obtain endpoint values. For
matrix-valued u, those return scalar array entries, not the first/last output column. This affects matrix-valued extrapolation more generally, not only Akima.

Because of that, matrix-valued extrapolation and extrapolated derivative/integral behavior should be addressed in a separate PR with a global fix.

## Checklist

## Additional information
- Fixes #468.
- This PR builds on the earlier work in #467. I reviewed that implementation and kept the scope focused on matrix-valued `AkimaInterpolation` construction, in-domain evaluation, derivatives, and integrals.

### DIFF
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -215,6 +215,13 @@ function _derivative(A::AkimaInterpolation{<:AbstractVector}, t::Number, iguess)
     return @evalpoly wj A.b[idx] 2A.c[j] 3A.d[j]
 end
 
+function _derivative(A::AkimaInterpolation{<:AbstractMatrix}, t::Number, iguess)
+    idx = get_idx(A, t, iguess; idx_shift = -1, side = :first)
+    j = min(idx, size(A.c, 2))  # for smooth derivative at A.t[end]
+    wj = t - A.t[idx]
+    return @evalpoly wj A.b[:, idx] 2A.c[:, j] 3A.d[:, j]
+end
+
 function _derivative(A::ConstantInterpolation, t::Number, iguess)
     return zero(first(A.u))
 end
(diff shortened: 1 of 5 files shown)


sathvikbhagavan:
Thank you for the PR! This was part of #577, so this feature is now present on master.

PR #525, opened 2026-05-15
Add modified Akima (makima) as a kwarg on AkimaInterpolation

ChrisRackauckas-Claude:
## Summary

Adds a `modified::Bool = false` keyword argument to `AkimaInterpolation` that switches the slope computation at the knots to the modified Akima (makima) formulation:

```julia
AkimaInterpolation(u, t; modified = true)
```

This is an alternative integration of what #524 proposed: rather than adding a separate set of `makima_*` helper functions / a parallel evaluation path, the same data is exposed as an option on the existing `AkimaInterpolation` type. The struct fields, evaluation, derivative, and integral methods are all unchanged — only the formula for the slope coefficients `b` differs when `modified = true`.

## Motivation

Akima and Makima share the same piecewise cubic representation:
\$\$ y(t) = u_i + b_i (t - t_i) + c_i (t - t_i)^2 + d_i (t - t_i)^3 \$\$
and differ only in how the per-knot slope \$b_i\$ is computed from the local divided differences:

- **Akima:** weights \$f_1 = |m_{i+1} - m_i|\$, \$f_2 = |m_{i-1} - m_{i-2}|\$
- **Makima:** weights \$w_1 = |m_{i+1} - m_i| + |m_{i+1} + m_i|/2\$, \$w_2 = |m_{i-1} - m_{i-2}| + |m_{i-1} + m_{i-2}|/2\$ (the extra term reduces overshoot on flat regions)

Because everything downstream of \$b, c, d\$ is identical, the cleanest representation is one type with a flag. This avoids:

- Duplicating `_interpolate`, `_derivative`, `_integral`, `cumulative_integral`, plotting, show, AD rules, allocation tests, etc.
- A new exported name for what is essentially the same interpolation family.

## What is in this PR

- `src/interpolation_caches.jl`: branch on `modified` in the `AkimaInterpolation` constructor. The standard Akima path is unchanged. The makima path has a simple-average fallback parallel to the standard Akima's `f12 > tol * max(f12)` guard, so the all-zero-slopes edge case (e.g. a step-shaped signal) does not divide by zero.
- `test/interpolation_tests.jl`: adds a `Modified Akima (makima)` sub-testset alongside the existing Akima tests, verifying:
  - Knot interpolation \$A(t_i) = u_i\$
  - Coefficients agree with a direct vectorized makima formula
  - Output differs from standard Akima on data where it should
  - The all-zero-slopes edge case (e.g. `u = [0,0,0,1,1,1]`) yields finite, correct values
  - Extrapolation, derivatives, and integrals all work through the shared struct path

## Test plan

- [ ] CI passes
- [x] `Pkg.test()` Core / Methods / Extensions / Misc groups all pass locally on Julia 1.10
- [x] Runic-formatted

Closes #524 as an alternative.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -17,6 +17,7 @@ makedocs(
     ),
     linkcheck_ignore = [
         "https://www.mathworks.com/content/dam/mathworks/mathworks-dot-com/moler/interp.pdf",
+        "https://blogs.mathworks.com/cleve/2019/04/29/makima-piecewise-cubic-interpolation/",
     ],
     pages = [
         "index.md",
--- a/src/interpolation_caches.jl
+++ b/src/interpolation_caches.jl
@@ -220,7 +220,7 @@ function LagrangeInterpolation(
 end
 
 """
-    AkimaInterpolation(u, t; extrapolation::ExtrapolationType.T = ExtrapolationType.None, extrapolation_left::ExtrapolationType.T = ExtrapolationType.None,
+    AkimaInterpolation(u, t; modified = false, extrapolation::ExtrapolationType.T = ExtrapolationType.None, extrapolation_left::ExtrapolationType.T = ExtrapolationType.None,
         extrapolation_right::ExtrapolationType.T = ExtrapolationType.None, cache_parameters = false)
 
 It is a spline interpolation built from cubic polynomials. It forms a continuously differentiable function. For more details, refer: [https://en.wikipedia.org/wiki/Akima_spline](https://en.wikipedia.org/wiki/Akima_spline).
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
Status after the latest push (3eca5317):

**Passing (14 jobs)**: spell check, runic, Tests Core × 3 (1/lts/pre), Tests Methods × 3, Tests Misc × 3, Tests Extensions (1 & lts).

**Failing (4 jobs, all pre-existing on master, not introduced by this PR)**:
- `Tests (1, QA)` and `Tests (lts, QA)` — `alloc_tests.jl :: LinearInterpolation` asserts `isempty(stderr_content)`, but Julia 1.10's LLVM emits a `lookup(jljit, ...)` deprecation warning. I verified this fails identically on a clean `upstream/master` checkout (reproduction: `ENV["GROUP"]="QA"; Pkg.test()` on master prints the same warning and fails the same assertion). Last green master `Tests` run was several days ago.
- `Tests (pre, Extensions)` — also failing on master `Tests` runs for the last 2-4 days.
- `Documentation` — was failing on the prior push because my docstring linked to a MathWorks blog URL that returns 403 to Documenter's linkchecker. Fixed in 3eca5317 by adding the URL to `linkcheck_ignore` alongside the existing `moler/interp.pdf` entry; that should clear once the docs job re-runs.

Happy to file the QA / pre-Extensions issues separately if useful, but they're unrelated to this PR and out of scope for it.

PR #531, opened 2026-05-21
v2: search_properties kwarg + get_idx via Auto + QuadraticSpline O(n) + FastInterpolations bench

ChrisRackauckas-Claude:
## Summary

Coordinated update with FindFirstFunctions.jl PR #73 (v3.0 release with enum-tagged dispatch + parametric `SearchProperties{T}` + props-aware `UniformStep`). This PR follows up on the merged #529.

The original #531 had six pieces:

1. **Drop the legacy `linear_lookup` / `seems_linear` / `looks_linear` machinery.** Every interpolation cache had a `linear_lookup::Bool` field set from a hand-rolled `looks_linear(t; threshold)` probe. That's exactly what `t_props.is_uniform` already gives us. **Breaking change**: `assume_linear_t` kwarg removed from all constructors. For explicit override, use `SearchProperties(t; is_uniform = true)`.

2. **Refactor `get_idx` to dispatch through `Auto(A.t_props)`.** (Now updated in this revision — see point 6 below.)

3. **Fix O(n²) QuadraticSpline construction.** `quadratic_spline_params` was calling `spline_coefficients!` which had a `findfirst(x -> x > u, k)` linear scan inside a per-knot loop. Replaced with a running pointer that advances monotonically → O(n) total. ~870× speedup at n=100k.

4. **`search_properties` constructor kwarg on every interpolation type.** Optional `search_properties::Union{Nothing, FindFirstFunctions.SearchProperties} = nothing`. When omitted, behaviour is identical to before; when supplied, the caller's `SearchProperties` is used as-is.

5. **FastInterpolations.jl added to the cross-library bench.** `bench/cross_library_comparison.jl` now benchmarks FastInterpolations.jl across Linear/Cubic/Quadratic/Akima/PCHIP; `bench/fast_interpolations_bench.jl` ports their advertised bench.

6. **Route `_resolve_strategy` through `Auto(t)`** (updated in this revision). The earlier draft pinned `_resolve_strategy(t)` to `BracketGallop()` to avoid Union-typed strategy fields. With FFF v3's parametric `Auto{T}`, the strategy field is now `Auto{T}` where `T` is the data ratio type — single concrete per `t`, type-stable. `Auto(t)` resolves to:
   - `KIND_UNIFORM_STEP` for uniformly-spaced data, taking the props-aware closed-form kernel (one subtract, one multiply, one truncate per query).
   - `KIND_LINEAR_SCAN` for `length(t) ≤ 16` non-uniform.
   - `KIND_BRACKET_GALLOP` otherwise (the previous default).

   Mooncake ext gains a `@zero_adjoint` declaration for `searchsortedlast(::Auto, ...)` — the integer index isn't differentiable and Mooncake's recursion through `Auto{Float64}` (which contains `Float64` `first_val` / `inv_step` fields) would otherwise hit `llvmcall` SIMD intrinsics in FFF's strategy kernels.

## Per-query latency (n = 10k, m = 1k, ns/query, BenchmarkTools median)

| Workload | DI before (BG pinned) | **DI after (Auto + props)** | FastInterpolations |
|---|---|---|---|
| Range, sorted queries | 89 | **75** (-16%) | 3.2 |
| Range, random queries | 89 | **76** (-15%) | 3.3 |
| Range, chained (monotone) | 89 | **75** (-16%) | 3.3 |
| Uniform Vec, sorted queries | 47 | **32** (-32%) | 70 |
| Uniform Vec, random queries | 50 | **35** (-30%) | 92 |
| Uniform Vec, chained | 48 | **32** (-33%) | n/a |
| Non-uniform Vec, sorted queries | 68 | 85 (+25%) | 75 |
| Non-uniform Vec, random queries | 80 | 95 (+19%) | 87 |
| Non-uniform Vec, chained | 67 | 86 (+28%) | n/a |

Wins: 15-33% on uniform data (Range and uniform `Vector`). Cost: ~5-20 ns/q on non-uniform Vector (Auto's per-query `s.kind === KIND_UNIFORM_STEP` branch adds register pressure on the BracketGallop path; FFF's enum-dispatch function body inlines both KIND paths into the call site).

The DI ↔ FastInterp gap on Range narrowed from 26× (89 ns/q ÷ 3.5 ns/q) to **22×** (75 ÷ 3.3). The remaining gap is DI's per-query overhead (Guesser hint, extrapolation check, linear-interp arithmetic), not the strategy. Closing further would require fusing search + interp + extrapolation into a single kernel.

Reproducer: `bench/di_perq_bench.jl` (added this revision).

## Tests

All 5 groups pass on Julia 1.12:
- **Core**: 17555 pass (5 pre-existing broken)
- **Methods**: 41417 pass
- **Extensions**: pass (SCT + Zygote + Mooncake; Mooncake required the `@zero_adjoint` fix above)
- **Misc**: 11 pass
- **QA** (Aqua + AllocCheck): 18 pass

## FFF v3 dependency

`[compat]` requires `FindFirstFunctions = "3"` (FFF #73 must merge and register first; CI here cannot resolve until then). v3 brings:
- Enum-tagged dispatch for singleton search strategies (zero-overhead vs v2 multimethod path).
- Parametric `SearchProperties{T}` with precomputed `first_val::T` / `inv_step::T`.
- Parametric `Auto{T}` carrying `SearchProperties{T}`.
- Props-aware `UniformStep` kernel folding in the closed-form O(1) lookup from #74 (which is now closed as superseded).

FFF v3 removed the v2 `Base.searchsortedlast(::S, ...)` extensions entirely; DI's call sites (`get_idx`, Mooncake `@zero_adjoint`, test helpers) are migrated to `FindFirstFunctions.search_last` / `search_first`.

## Notes

Draft. Please ignore until reviewed by @ChrisRackauckas.

Co-Authored-By: Chris Rackauckas <accounts@chrisrackauckas.com>


ChrisRackauckas-Claude:
## Update: rebased + correctness fixes + verification (14bda8ba)

Rebased onto current `master` (Mooncake 0.5 bump, RegularizationTools removal, canonical CI). Conflicts were confined to `Project.toml` compat; upstream's tightened floors were kept and `FindFirstFunctions = "3"` applied.

### Fixes in this push

- **Uniform fast path hardening (`ff4eb140`).** Two issues in the new statically-dispatched LinearInterpolation kernel: (1) the closed-form float position was `unsafe_trunc`'d before clamping — UB for queries far outside the knot span, reachable with `ExtrapolationType.Extension` (returned garbage for `t = 1e300`); (2) `push!`/`append!` mutate `A.t` while `t_props` and the `IsUniform` type tag keep construction-time values, so a uniformity-breaking push silently corrupted values. The kernel now clamps in the float domain and verifies its guessed cell + spacing against the live knots, falling back to the (extracted) slope-form path on mismatch. AllocCheck still passes. Regression tests added for both, plus a knot vector crafted to fool FFF's sampled uniformity probe (rejected since SciML/FindFirstFunctions.jl#73's exact-validation fix).
- **Strategy/props consistency (`0e8d57cd`).** Constructors computed `t_props` and then called `Auto(t)`, which re-probed and ignored the cached/user-supplied props — redundant O(n) scan and `A.strategy.props` could disagree with `A.t_props`. Now resolved via `Auto(t, t_props)`.
- **Docs (`980c1bcb`, `14bda8ba`).** Removed the stale `looks_linear` `@docs` entry (would fail the docs build), documented the breaking `assume_linear_t` removal and the new features in NEWS, added the `search_properties` kwarg bullet to all 16 constructor docstrings, and restored Vector-knot constructor `@inferred` coverage for the types that don't carry the `IsUniform` tag.

### Independent verification (subagent, scripts vs merge-base worktree)

- **QuadraticSpline O(n) rewrite:** `(k, A)` elementwise identical to the legacy `findfirst` loop on 25 edge cases (duplicates, Float32/Int/Rational, n=2/3); off-knot values bit-identical PR vs merge-base; locator parity on 200 random cases. Three edge-case crashes found are pre-existing on master → filed #542.
- **Mooncake ext:** full `mooncake_tests.jl` 604/604 on Mooncake 0.5.31; t-gradients match analytic derivatives < 1e-8 on uniform (`KIND_UNIFORM_STEP`) and non-uniform knots; the new `increment_and_get_rdata!` method is provably required (deleting it reproduces the fdata-type error); zero-adjoint does not cut genuine derivative flow.
- **Online mutation:** push!/append! breaking uniformity on Linear/Constant/Quadratic × cache_parameters — value/derivative/integral exact vs freshly-constructed interpolations (the FFF props kernel walk-corrects against live knots; the DI fast path falls back via verification).

### Tests (local, Julia 1.12, FFF#73 dev'd)

| Group | Result |
|---|---|
| Core | 17,555 pass (5 pre-existing broken) |
| Methods | 41,417 pass |
| Extensions | 13,178 pass |
| Misc | 10 pass |
| QA | 18 pass (Aqua + AllocCheck) |

Runic clean. Note for merge ordering: this PR requires FindFirstFunctions v3.0.0 (#73) to be merged and registered first — CI here will fail to resolve until then.

ChrisRackauckas-Claude:
## Update: migrated to FFF v3's shim-free API (`1ab2ed1c`)

SciML/FindFirstFunctions.jl#73 now removes the v2 `Base.searchsortedlast(::S, ...)` / `Base.searchsortedfirst(::S, ...)` extensions outright (breaking release, no compat layer). DI's call sites are migrated in lockstep:

- `get_idx` calls `FindFirstFunctions.search_last` / `search_first` on the cached `Auto` strategy.
- The Mooncake `@zero_adjoint` declarations target `typeof(FindFirstFunctions.search_last/search_first)` instead of the removed Base methods.
- Test helpers (`test_cached_index`, the derivative cached-index check) use the qualified v3 names.

**All 5 groups pass locally on Julia 1.12 against the updated FFF branch:** Core 17,555 (5 pre-existing broken) / Methods 41,417 / Extensions 13,178 (incl. Mooncake + Zygote + SCT) / Misc 10 / QA 18. Runic clean.

ChrisRackauckas-Claude:
Follow-up to the FFF rename (`150dfb38`): FindFirstFunctions renamed its dispatchers `search_last`/`search_first` → `searchsorted_last`/`searchsorted_first` to keep the "sorted" precondition in the name. DI's call sites (`get_idx`, the Mooncake `@zero_adjoint` declarations, and the test helpers) updated to match. All 5 groups still pass locally on Julia 1.12: Core 17,555 / Methods 41,417 / Extensions 13,178 / Misc 10 / QA 18. Runic clean.

ChrisRackauckas-Claude:
## FindFirstFunctions v3.0.0 is registered — this is now resolvable

Rebased onto current `master` (picks up #543, the QuadraticSpline edge-case fix for the issues I'd filed in #542). No conflict, but both that PR and this one touch `quadratic_spline_params`; I verified the merged function carries **both** changes — this PR's O(n) running-locator loop *and* #543's duplicate-knot `ArgumentError` guard + `dtype_sc = typeof(one(eltype(t)) / one(eltype(t)))` fix. The two test sets coexist.

Validated against the **registered** `FindFirstFunctions v3.0.0` (resolved from General, git-tree `141f3cfb` — byte-identical to merged #73), not a local dev copy, so this mirrors CI exactly:

| Group | Result |
|---|---|
| Core | (running) |
| Methods | (running) |
| Extensions | (running) |
| Misc | (running) |
| QA | (running) |

Compat is `FindFirstFunctions = "3"`; the earlier "register v3 first" caveat no longer applies. Runic clean. Will confirm the group results once the local run finishes.

ChrisRackauckas-Claude:
Full suite passes against the **registered** FindFirstFunctions v3.0.0 (resolved from General, not a local dev copy):

| Group | Result |
|---|---|
| Core | 17,566 pass (5 pre-existing broken) |
| Methods | 41,417 pass |
| Extensions | 13,178 pass (Mooncake + Zygote + SCT) |
| Misc | 10 pass |
| QA | 18 pass (Aqua + AllocCheck) |

Also audited the API surface statically: all six `FindFirstFunctions` symbols this PR references (`Auto`, `SearchProperties`, `searchsorted_last`/`searchsorted_first`, `searchsortedlast!`, `SearchStrategy`) exist in the released v3; every call pattern and `t_props`/`Auto` field this PR reads resolves; and there are no leftover references to the removed v2 API (`Base.searchsorted*` strategy methods, the old `search_last`/`search_first` names, or `assume_linear_t`/`looks_linear`) anywhere in src/ext/test/docs. Ready for review.

ChrisRackauckas-Claude:
## CI triage — DI's own changes are complete; remaining reds are not DI defects

All **test** groups pass on CI across Julia 1/lts/pre (Core, Methods, Extensions, Misc, QA). Three non-test jobs are red; here's the disposition:

- **Spell Check** — *mine, fixed* (`efcacdf7`). The `typos` checker flagged my local variable `strat` in `get_idx` as a misspelling of `start`; renamed to `strategy`. (Spell Check is green on master, so this was the only regression I introduced.)
- **Documentation** — *ecosystem lag, not a DI defect.* The docs environment pulls ModelingToolkit/OrdinaryDiffEq for examples; MTK depends on `StateSelection`, which still caps `FindFirstFunctions = "1.2.0 - 2"`. Once this PR requires FFF v3, the docs env can't resolve (`StateSelection ... restricted by FindFirstFunctions to versions: uninstalled`). This is the expected first-consumer-of-a-breaking-release situation. StateSelection's only FFF usage is `findfirstequal`, which v3 keeps unchanged, so it's code-safe to widen its compat to `"…, 3"` — but it's a JuliaComputing repo, so flagging for @ChrisRackauckas rather than PRing it. The DI package, test env, and QA all resolve against registered FFF v3.0.0 cleanly.
- **Downgrade** — *pre-existing master failure, not from this PR.* The Downgrade workflow is red on every master commit since ~June 13 (last green June 8 at `bf7c4edd`, this PR's merge-base) — a `--min=@deps` floor-resolution `Unsatisfiable`. I've spawned an investigation to bisect the master regression and name the conflicting floor; will report separately. It is independent of the FFF v3 change.

Net: nothing further is needed in DI itself for the v3 that landed. Will confirm Spell Check goes green and report the Downgrade root cause.

ChrisRackauckas-Claude:
Rebased onto current `master` (`4d74209f`), picking up #545/#546 (the Optim/Mooncake/StaticArrays compat-floor bumps) and **#547** (the SciMLTesting v1.2 folder-based test harness). No conflicts; the rebase correctly followed #547's file moves — my `derivative_tests.jl` edit moved to `test/Methods/`, the `interpolation_tests.jl`/`interface.jl` edits stayed in Core, no orphaned copies. Compat now carries `FindFirstFunctions = "3"` alongside master's floor bumps.

Re-verified under the new harness against the **registered** FindFirstFunctions v3.0.0: Core (interpolation_tests.jl 17,412 + my fast-path parity / jittered-probe / Extension-extrapolation / push!-staleness tests, plus #543's duplicate-knot test) and Methods (incl. the migrated `derivative_tests.jl`) both pass. Runic clean.

Note: master's Downgrade is still red from the Julia-1.10 floor conflict (#546's `Mooncake = "0.4.175"` floor + the Optim/Symbolics/Mooncake triple) — pre-existing and independent of this PR, as detailed above.

ChrisRackauckas-Claude:
## Collapsed the redundant strategy field (`8e654164`)

Per review feedback: every cache stored **both** `t_props::SearchProperties{T}` and `strategy::Auto{T}`, where `Auto = (kind, props)` — so the `SearchProperties` was duplicated in the struct and the cache carried two T-driven type parameters (`propsType` + `strategyType`) that always moved together.

Since the cached `strategy` was only ever read by `get_idx` (the fast path and batched eval already work off `t_props`, and the batched API re-resolves the kind anyway), I replaced `strategy::Auto{T}` with **`kind::StrategyKind`** — a `UInt8` enum, non-parametric — dropping the `strategyType` parameter from all 15 cache structs and the two integral-inverse caches. The parametric props payload remains as the single `t_props` field.

`get_idx` branches on the cached `kind`: `KIND_UNIFORM_STEP` reconstructs the (isbits, stack-allocated) `Auto` from `t_props` so the closed-form O(1) lookup is taken; every other kind dispatches the bare enum, preserving the hint-aware gallop. The Mooncake ext gains `@zero_adjoint` declarations for the `StrategyKind` dispatch form.

So the answer to "isn't the strategy an enum?" is now yes — it's stored as the enum; the only remaining parametric field is `t_props`, which is irreducible (it holds the `first_val::T`/`inv_step::T` reciprocals the uniform kernel needs).

Re-verified: all 5 groups pass against **registered FindFirstFunctions v3.0.0** under the new SciMLTesting harness — Core, Methods, Extensions (Mooncake/Zygote/SCT — validates the new `StrategyKind` zero-adjoints), Misc, and QA (AllocCheck confirms the `Auto` reconstruction is allocation-free). Runic clean.

ChrisRackauckas-Claude:
## Fixed the inference regression — uniformity is now a runtime enum, not a type parameter (`50195f85`)

The static fast-path work had encoded knot uniformity in a `Val{IsUniform}` **type parameter** on `LinearInterpolation`. Because a `Vector`'s uniformity is a runtime property, the constructor returned `Union{LinearInterpolation{…,true}, LinearInterpolation{…,false}}` — type-unstable — and that had been papered over by relaxing/removing the `@inferred` tests. Not acceptable.

Fixed by following FindFirstFunctions' own design (runtime enum dispatch → concrete inferred returns):

- **Dropped the `IsUniform` type parameter** and the `is_uniform_static::Val` field. `LinearInterpolation` is a single concrete type again, so the **constructor is inferred for both Vector and Range knots** (it was returning an abstract `Union` before).
- **The uniform closed-form path is now selected by a runtime branch** on the cached `kind` enum inside `_interpolate` (`A.kind === KIND_UNIFORM_STEP`), both arms returning the same concrete type so the **query stays inferred**. Same body as before — just reached by a runtime branch instead of type dispatch, exactly like FFF's `StrategyKind` dispatch returns a concrete `Int` regardless of the runtime kind.
- Removed the now-unused `_static_uniform_tag`.
- **Restored the inference tests**: the `Type Inference` testset `@inferred`s every constructor for Vector *and* Range knots (plus query inference), and the `LinearInterpolation` testset's relaxed `@inferred` guards are reverted to master's unconditional form.

Verified: all 7 constructors and all 13 interpolation query paths `@inferred`-clean; full suite green against registered FFF v3.0.0 — Core (incl. the restored Type Inference testset, 28 passing), Methods, Extensions (Mooncake 604 / Zygote 11974 / SCT 600 — exercises AD through the runtime-branch fast path), Misc, QA (AllocCheck 8 — the `Auto`/fast-path reconstruction stays allocation-free). Runic clean.

ChrisRackauckas-Claude:
## Performance: static lean fast path for `AbstractRange` knots (`f09c1b89`)

Following up on the inference fix — I benchmarked whether the runtime `kind` branch costs anything, and whether a static/runtime split is warranted (it is).

**Findings (per-query, n=10k, monomorphic loop):**

| Path | Range knots | Uniform Vector |
|---|---|---|
| runtime-branch `A(q)` (before) | 30.5 ns/q | 18.4 ns/q |
| **+ static range fast path (after)** | **17.0 ns/q** | 18.4 ns/q (unchanged) |

The runtime branch itself is ~free on Vectors (≈0.5 ns). The real opportunity was that a **range was paying the *Vector* kernel's overhead** — cell/spacing verification, two `A.t[idx]` range-arithmetic loads, and the iguesser store — all provably unnecessary for an immutable, exactly-uniform range.

**The split (two inference-safe uniform paths):**
- **Static**, for `t::AbstractRange`: dispatched on `typeof(t)` (inference-safe — no value-dependent `Union`), a lean kernel that skips the branch + verification + `A.t` loads (`α = f - idx0`) + the vestigial iguesser store. Keeps full NaN handling.
- **Runtime**, for `Vector` knots: the `kind`-branch verified kernel — needed because a `Vector` can be `push!`-mutated or caller-forced-uniform on jittered data. (`search_properties = SearchProperties(t; is_uniform=true)` designates uniformity via the runtime `kind`; a Vector type can't carry a static guarantee.)
- Non-uniform / non-Float `u`: slope form.

**Verified:** all 5 groups green against registered FFF v3.0.0; inference clean for both Range and Vector (constructor + query); AD correct through the new kernel (Mooncake 604 / Zygote 11974 / SCT 600); allocation-free (AllocCheck). Range interior matches the slope form to 3e-15.

ChrisRackauckas (approved):

PR #533, opened 2026-05-29
Fix data race in BSpline and QuadraticSpline evaluation

### LINKED ISSUE #532: Data Race with BSplines
```julia
# ============================================================================
#  Data race in `BSplineInterpolation` under concurrent evaluation
# ============================================================================
#
#  WHAT:  Evaluating a single, shared `BSplineInterpolation` from several
#         threads silently returns WRONG values (no error is thrown).
#
#  WHY:   `BSplineInterpolation` keeps a preallocated coefficient buffer in its
#         `sc` field and OVERWRITES it in place on every evaluation
#         (src/interpolation_methods.jl):
#
#             # reuse the shared buffer A.sc (only the Dual branch allocates fresh)
#             sc = t isa ForwardDiff.Dual ? zeros(eltype(t), n) : A.sc
#             spline_coefficients!(sc, A.d, A.k, t)   # writes basis(t) into A.sc
#             for i in nonzero_coefficient_idxs
#                 ucum += sc[i] * A.c[i]              # then reads A.sc back
#             end
#
#         So a "read" (evaluation) mutates state inside the object, and it is
#         not reentrant. Two threads at different points race on `A.sc`: one
#         thread's `spline_coefficients!` overwrites the buffer between the
#         other thread's write and read, producing a garbage result.
#
#  FIX:   Take the same fresh-buffer path the `ForwardDiff.Dual` branch already
#         uses for the plain-number case (`sc = zeros(eltype(t), n)`), or
#         document `BSplineInterpolation` as non-reentrant for concurrent eval.
#
#  NOTE:  Local interpolants (`CubicSpline`, `LinearInterpolation`, PCHIP, ...)
#         recompute each segment's parameters from read-only fields per call and
#         keep no such scratch buffer, so they are unaffected — see Part 3.
#
#  RUN:   julia -t auto di_threadsafety_mwe.jl
# ----------------------------------------------------------------------------

using DataInterpolations
using Base.Threads

println("DataInterpolations v", pkgversion(DataInterpolations),
        " | Julia ", VERSION, " | threads = ", nthreads())

# Minimal toy data — the bug is general, nothing here is domain-specific.
# The two probe points (2.5, 5.5) have very different values, so a corrupted
# result is unmistakable rather than off in the last bit.
u = [0.0, 1.0, 0.0, 1.0, 0.0, 1.0]
t = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]
itp = BSplineInterpolation(u, t, 3, :Uniform, :Average)

# ── Part 1 — the root cause, with NO concurrency ────────────────────────────
# Evaluate at one point, then another: the SAME internal buffer `itp.sc`
# changes. In other words, evaluating the curve mutates hidden state.
itp(2.5); buffer_after_2_5 = copy(itp.sc)
itp(5.5); buffer_after_5_5 = copy(itp.sc)
println("\n[1] itp.sc is one buffer, overwritten on every evaluation: ",
        buffer_after_2_5 != buffer_after_5_5)              # => true

# ── Part 2 — the consequence under threads ──────────────────────────────────
# Each point has one correct value (computed single-threaded just below). A
# thread-safe curve would always return it. Evaluating the two points
# concurrently, the shared buffer gets clobbered and some results are wrong.
@assert nthreads() > 1 "Start Julia with >1 thread, e.g. `julia -t auto`."
correct_2_5, correct_5_5 = itp(2.5), itp(5.5)
wrong = Atomic{Int}(0)
@threads for _ in 1:1_000_000
    itp(2.5) == correct_2_5 || atomic_add!(wrong, 1)
    itp(5.5) == correct_5_5 || atomic_add!(wrong, 1)
end
println("[2] BSplineInterpolation: ", wrong[], " wrong / 2,000,000 evaluations")  # thousands

# ── Part 3 — control: a local interpolant keeps no scratch buffer ───────────
safe = CubicSpline(u, t)
ok_2_5, ok_5_5 = safe(2.5), safe(5.5)
wrong_safe = Atomic{Int}(0)
@threads for _ in 1:1_000_000
    safe(2.5) == ok_2_5 || atomic_add!(wrong_safe, 1)
    safe(5.5) == ok_5_5 || atomic_add!(wrong_safe, 1)
end
println("[3] CubicSpline (control): ", wrong_safe[], " wrong / 2,000,000 evaluations")  # 0
```

**Error & Stacktrace ⚠️**

```Julia
# running the above produces
DataInterpolations v8.10.0 | Julia 1.12.5 | threads = 10

[1] itp.sc is one buffer, overwritten on every evaluation: true
[2] BSplineInterpolation: 1591 wrong / 2,000,000 evaluations
[3] CubicSpline (control): 0 wrong / 2,000,000 evaluations

```

alecloudenback:
Fixes #532.

## Problem

`BSplineInterpolation`, `BSplineApprox`, and `QuadraticSpline` (with the default `cache_parameters = false`) keep a scratch buffer in their `sc` field and overwrite it in place on **every evaluation**:

```julia
sc = t isa ForwardDiff.Dual ? zeros(eltype(t), n) : A.sc
spline_coefficients!(sc, A.d, A.k, t)   # writes basis(t) into A.sc
for i in nonzero_coefficient_idxs
    ucum += sc[i] * A.c[i]              # then reads A.sc back
end
```

Because a "read" (evaluation) mutates state inside the object, evaluation is **not reentrant**. Two threads evaluating a single shared interpolant race on `A.sc`: one thread's `spline_coefficients!` overwrites the buffer between the other thread's write and read, silently producing a wrong result (no error is thrown).

The issue reported `BSplineInterpolation`, but the same root cause affects `QuadraticSpline` via `quadratic_spline_parameters` (and therefore its derivatives and integrals too). Confirmed on this branch's pre-fix state: `QuadraticSpline` gives ~12k wrong / 2M concurrent evals; `BSplineInterpolation` ~1.5–4k wrong / 2M.

## Fix

- **`BSplineInterpolation` / `BSplineApprox`** (`interpolation_methods.jl`, `derivatives.jl`): allocate the coefficient scratch per call instead of reusing `A.sc`. This is exactly the fix suggested in the issue and collapses into the pre-existing `ForwardDiff.Dual` branch.
- **`QuadraticSpline`** (`parameter_caches.jl`): evaluate the three nonzero degree-2 B-spline basis values in local variables instead of writing them into the shared `A.sc` buffer. Numerically identical to the old `spline_coefficients!` path (verified to `0.0` over 5000 random knot configs), and **allocation-free** for both scalar and `SVector`-valued data, so the `alloc_tests.jl` guarantee is preserved.
- **Regression tests** (`test/thread_safety_tests.jl`, wired into the Methods group): a deterministic "evaluation does not mutate the interpolant" check that catches the bug even on a single thread, plus a concurrent stress test that runs when more than one thread is available. Covers scalar and array-valued `BSplineInterpolation`/`BSplineApprox`, a degree-1 (linear) `BSplineInterpolation`, and `QuadraticSpline` (cached + uncached).

## Linear B-splines

For clarity, since it came up: a degree-1 `BSplineInterpolation` ("linear B-spline") **is** affected and **is** fixed — it dispatches to the same `_interpolate` method as higher degrees, so it shared the race (35 wrong / 1M before → 0 after). The separate `LinearInterpolation` type is **not** affected: it keeps no scratch buffer (no `sc` field) and recomputes from read-only fields per call, exactly like the `CubicSpline` control in the issue.

## Benchmarks

Apple silicon, Julia 1.12.5, `@belapsed`, single interior evaluation point; `n` = number of data points (BSpline scratch buffer is size `n`, BSplineApprox size `h = n÷2`).

**Evaluation `A(x)` — time / allocations:**

| type | n | before | after |
|---|---|---|---|
| `BSplineInterpolation` | 10 | 58 ns / 0 B | 48 ns / 144 B |
| | 100 | 74 ns / 0 B | 77 ns / 928 B |
| | 1000 | 263 ns / 0 B | 381 ns / 8256 B |
| `BSplineApprox` | 10 | 60 ns / 0 B | 44 ns / 96 B |
| | 100 | 71 ns / 0 B | 67 ns / 480 B |
| | 1000 | 241 ns / 0 B | 332 ns / 4160 B |
| `QuadraticSpline` (uncached, default) | 10 | 59 ns / 0 B | **19 ns / 0 B** |
| | 100 | 75 ns / 0 B | **26 ns / 0 B** |
| | 1000 | 271 ns / 0 B | **145 ns / 0 B** |
| `QuadraticSpline` (`cache_parameters=true`) | 10–1000 | ~16 ns / 0 B | ~17 ns / 0 B |

**Derivative `derivative(A, x)` (BSpline only):**

| type | n | before | after |
|---|---|---|---|
| `BSplineInterpolation` | 10 / 100 / 1000 | 31 / 80 / 725 ns, 0 B | 37 / 98 / 853 ns, 144 / 928 / 8256 B |
| `BSplineApprox` | 10 / 100 / 1000 | 29 / 55 / 456 ns, 0 B | 34 / 67 / 559 ns, 96 / 480 / 4160 B |

Takeaways:
- **`QuadraticSpline` (the default) gets ~2–3× faster and stays allocation-free.** The old path called `spline_coefficients!`, which zeroed the entire size-`n` buffer (`N .= 0`) on every call; the new functional basis computes only the 3 nonzero values.
- **`BSpline*` now allocates one scratch buffer per call** (it was allocation-free). Time impact is negligible at small `n` and grows with the buffer at large `n` (e.g. eval +45% at `n=1000`, where the 8 KB allocation dominates).
- `QuadraticSpline` with `cache_parameters=true` is unchanged.

## Tradeoff

`BSpline*` evaluation now allocates a small per-call buffer where it previously reused `A.sc`. With a runtime-variable spline degree this is the straightforward cost of reentrancy, and it matches the fix proposed in the issue; BSpline eval is not part of the allocation-free test suite. `QuadraticSpline` (fixed degree 2) stays allocation-free — and is now faster. If preferred, I can make `BSpline*` allocation-free as well via a small stack buffer (bounded max degree with a heap fallback) — happy to follow up.

Note: the `sc` struct fields are now unused at evaluation time (still built at construction). I left them in place to keep this PR focused on the correctness fix; removing them touches the struct definitions, type parameters, and constructors and can be a separate cleanup.

### DIFF
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -256,7 +256,8 @@ function _derivative(A::BSplineInterpolation{<:AbstractVector{<:Number}}, t::Num
     n = length(A.t)
     scale = (A.p[idx + 1] - A.p[idx]) / (A.t[idx + 1] - A.t[idx])
     t_ = A.p[idx] + (t - A.t[idx]) * scale
-    sc = t isa ForwardDiff.Dual ? zeros(eltype(t), n) : A.sc
+    # Per-call scratch buffer: evaluation must be reentrant for thread safety (#532)
+    sc = zeros(eltype(t), n)
     spline_coefficients!(sc, A.d - 1, A.k, t_)
     ducum = zero(eltype(A.u))
     if t == A.t[1]
@@ -280,7 +281,8 @@ function _derivative(
     n = length(A.t)
     scale = (A.p[idx + 1] - A.p[idx]) / (A.t[idx + 1] - A.t[idx])
     t_ = A.p[idx] + (t - A.t[idx]) * scale
-    sc = t isa ForwardDiff.Dual ? zeros(eltype(t), n) : A.sc
+    # Per-call scratch buffer: evaluation must be reentrant for thread safety (#532)
+    sc = zeros(eltype(t), n)
     spline_coefficients!(sc, A.d - 1, A.k, t_)
     ducum = zeros(size(A.u)[1:(end - 1)]...)
     if t == A.t[1]
@@ -302,7 +304,8 @@ function _derivative(A::BSplineApprox{<:AbstractVector{<:Number}}, t::Number, ig
     idx = get_idx(A, t, iguess)
     scale = (A.p[idx + 1] - A.p[idx]) / (A.t[idx + 1] - A.t[idx])
     t_ = A.p[idx] + (t - A.t[idx]) * scale
-    sc = t isa ForwardDiff.Dual ? zeros(eltype(t), A.h) : A.sc
+    # Per-call scratch buffer: evaluation must be reentrant for thread safety (#532)
+    sc = zeros(eltype(t), A.h)
     spline_coefficients!(sc, A.d - 1, A.k, t_)
     ducum = zero(eltype(A.u))
     if t == A.t[1]
@@ -325,7 +328,8 @@ function _derivative(
     idx = get_idx(A, t, iguess)
     scale = (A.p[idx + 1] - A.p[idx]) / (A.t[idx + 1] - A.t[idx])
     t_ = A.p[idx] + (t - A.t[idx]) * scale
-    sc = t isa ForwardDiff.Dual ? zeros(eltype(t), A.h) : A.sc
+    # Per-call scratch buffer: evaluation must be reentrant for thread safety (#532)
+    sc = zeros(eltype(t), A.h)
     spline_coefficients!(sc, A.d - 1, A.k, t_)
     ducum = zeros(size(A.u)[1:(end - 1)]...)
     if t == A.t[1]
--- a/src/interpolation_caches.jl
+++ b/src/interpolation_caches.jl
@@ -646,7 +646,7 @@ function QuadraticSpline(
     k, A = quadratic_spline_params(t, sc)
     c = A \ u
 
-    p = QuadraticSplineParameterCache(u, t, k, c, sc, cache_parameters)
+    p = QuadraticSplineParameterCache(u, t, k, c, cache_parameters)
     A = QuadraticSpline(
         u, t, nothing, p, k, c, sc, extrapolation_left,
         extrapolation_right, cache_parameters, assume_linear_t
@@ -686,7 +686,7 @@ function QuadraticSpline(
         end
     end
 
-    p = QuadraticSplineParameterCache(u, t, k, c, sc, cache_parameters)
+    p = QuadraticSplineParameterCache(u, t, k, c, cache_parameters)
     A = QuadraticSpline(
         u, t, nothing, p, k, c, sc, extrapolation_left,
         extrapolation_right, cache_parameters, assume_linear_t
--- a/src/interpolation_methods.jl
+++ b/src/interpolation_methods.jl
@@ -1151,7 +1151,8 @@ function _interpolate(
     idx = get_idx(A, t, iguess)
     t = A.p[idx] + (t - A.t[idx]) / (A.t[idx + 1] - A.t[idx]) * (A.p[idx + 1] - A.p[idx])
     n = length(A.t)
-    sc = t isa ForwardDiff.Dual ? zeros(eltype(t), n) : A.sc
+    # Per-call scratch buffer: evaluation must be reentrant for thread safety (#532)
+    sc = zeros(eltype(t), n)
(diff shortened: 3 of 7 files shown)


ChrisRackauckas:
Changed to a non-allocating version https://github.com/SciML/DataInterpolations.jl/pull/550

ChrisRackauckas:
But thanks, this made the final part easy to do: it's effectively your fix but just with MVectors.

PR #537, opened 2026-06-04
Bump Mooncake to 0.5 and docs OrdinaryDiffEq to 7 (completes #536)

ChrisRackauckas-Claude:
**Please ignore until reviewed by @ChrisRackauckas.** Draft.

## Summary

This is the dependabot compat bump from #536 (Mooncake `0.4` → `0.4, 0.5`, OrdinaryDiffEq `6` → `6, 7.0` in `docs/`) **plus** a fix for the test-resolution failure that made every test group fail on that branch (and on `master`).

It is opened from a fork branch because `ChrisRackauckas-Claude` does not currently have push access to `SciML/DataInterpolations.jl` (direct `git push` returns 403; the collaborator role reports no `push` permission). Re-target / re-home as you prefer.

## Root cause (per failing group)

All five groups — **Core, Methods, Extensions, Misc, QA** — failed at the **same point**: the test harness re-resolve step (`Pkg.test(...; force_latest_compatible_version=true)`), with:

```
ERROR: LoadError: Unsatisfiable requirements detected for package RegularizationTools [29dad682]
... restricted by compatibility requirements with Optim [429524aa] to versions: uninstalled — no versions left
Optim ... restricted to versions 2 by an explicit requirement
```

- `force_latest_compatible_version=true` forces every test dep to its newest compatible version.
- The package declared `Optim = "1.7.5, 2"`, so it pins **Optim 2.x** (2.0.0 was registered Jan 2026; 2.1.0 May 2026).
- The latest **RegularizationTools 0.6.0** (an optional extension we test) declares `Optim = "0.18-0.22, 1"` — it does **not** support Optim 2, and there is no newer RegularizationTools release.
- Optim 2 and RegularizationTools 0.6 are therefore mutually exclusive → resolution fails before any test runs.

This is **not** caused by the Mooncake 0.5 / OrdinaryDiffEq bump itself — it is a latent conflict that surfaced once Optim 2 was published. I verified it **reproduces on a clean `master`** under the same force-latest resolve (master CI only passed earlier thanks to a cached manifest that kept Optim at 1.x; the dependabot branch had no such cache, so every group hit the clean re-resolve). `force_latest_compatible_version=true` is the relevant difference, not Mooncake.

OrdinaryDiffEq `7.0` is only in `docs/Project.toml` and does not affect the test groups.

## Fix

Cap `Optim = "1.7.5"` (drop the `2`). This:
- keeps the Optim `Curvefit` extension fully working — it uses only stable Optim 1.x API (`optimize`, `OnceDifferentiable`, `Fminbox`, `Optim.minimizer`);
- makes the force-latest test resolution satisfiable (selects **Optim 1.13.3** + **RegularizationTools 0.6.0**);
- leaves the downgrade floor unchanged (`1.7.5` was already the floor).

Trade-off: this removes Optim 2 support from the package until RegularizationTools supports Optim 2. A comment in `Project.toml` notes to re-add `2` then. (There is no per-test-env compat in this layout — `[compat]` governs both the package and the test extras — and a `test/Project.toml` is not used here.)

The Mooncake bump itself is fine: with **Mooncake 0.5.30** installed, `test/mooncake_tests.jl` (which compares Mooncake gradients against ForwardDiff and analytic `derivative`) passes in full, so the Mooncake 0.5 `friendly_tangents` default change did not break `DataInterpolationsMooncakeExt`.

## Local verification (Julia 1.11.9, `force_latest_compatible_version=true`)

Resolved versions: Mooncake **0.5.30**, Optim **1.13.3**, RegularizationTools **0.6.0**.

```
Test Summary: | Pass  Broken  Total     Time
Core          | 2539       5   2544  4m08.6s     # 5 Broken = pre-existing @test_broken
Test Summary: |  Pass  Total     Time
Methods       | 42151  42151  4m33.9s
Test Summary: |  Pass  Total      Time
Extensions    | 13178  13178  13m11.6s            # includes mooncake_tests.jl on Mooncake 0.5.30
Test Summary: | Pass  Total  Time
Misc          |   11     11  2.8s
     Testing DataInterpolations tests passed

# QA group (run separately):
Test Summary:     | Pass  Total   Time
Quality Assurance |   10     10  56.9s
Test Summary:    | Pass  Total   Time
Allocation Tests |    8      8  50.0s
     Testing DataInterpolations tests passed
```

No tests were skipped, disabled, or loosened. The only change is the `Optim` compat cap (plus the dependabot bump it sits on top of). Runic: clean (35/35 `.jl` files unchanged; the edit is TOML only).

Closes the breakage in #536.


ChrisRackauckas-Claude:
Rebased onto current `master`.

Context: this PR originally also capped `Optim` to `1.7.5` to work around RegularizationTools (the test-only extension) not supporting Optim 2. That workaround is now **obsolete** — #538 removed the deprecated RegularizationTools extension and restored `Optim = "1.7.5, 2"` on master. I dropped the cap commit (keeping it would revert #538) and rebased so this PR now carries **only** the two still-missing #536 compat bumps that master lacks:

- `Project.toml`: `Mooncake = "0.4"` → `"0.4, 0.5"`
- `docs/Project.toml`: `OrdinaryDiffEq = "6"` → `"6, 7.0"`

`Optim` stays at master's `"1.7.5, 2"`. Diff vs master is now those two lines only (ahead 1 / behind 0). Mooncake 0.5 was verified gradient-correct in the original pass; CI here will confirm resolution now that the RegularizationTools/Optim conflict is gone.

PR #550, opened 2026-06-28
Fix data race in BSpline and QuadraticSpline evaluation (rebase of #533)

ChrisRackauckas-Claude:
**Please ignore until reviewed by @ChrisRackauckas.**

Rebased version of #533 (original by @alecloudenback) onto current `master`. That PR was 18 commits behind and conflicting; this branch resolves the conflicts and updates it for the new test layout. Opened from the bot fork because I lacked push access to update #533's head branch directly. Supersedes #533 — close whichever is not merged.

## Original change (#532 fix)
`BSplineInterpolation`, `BSplineApprox`, and uncached `QuadraticSpline` kept a scratch buffer in their `sc` field and overwrote it in place on every evaluation, so a "read" mutated shared state — evaluating one shared interpolant from multiple threads raced and silently returned wrong values.

- BSpline{Interpolation,Approx}: allocate the spline-coefficient scratch per call instead of reusing `A.sc`, making evaluation reentrant.
- `QuadraticSpline`: compute the three nonzero degree-2 B-spline basis values in local variables instead of writing to the shared buffer.
- Thread-safety regression tests: a deterministic no-mutation-on-eval guard plus a concurrent stress test (now also covering degree-1 BSpline).

## Rebase conflict resolution
- **`src/parameter_caches.jl`**: master's #543 added a `length(t) == 2` edge-case guard while still using the shared `sc` buffer; #533 removed that buffer for thread safety. Merged both — kept the 2-point edge case **and** the reentrant local-variable basis evaluation.
- **`test/runtests.jl`**: master migrated to `SciMLTesting`'s folder-based `run_tests()` (#547), so the manual `@safetestset` list is gone. Took master's version and moved the new test to `test/Methods/thread_safety_tests.jl` so it's auto-discovered under the `Methods` group.

## Verification (run locally)
- All 28 thread-safety tests pass with `julia -t4` (concurrent stress path exercised, not skipped).
- 2-point `QuadraticSpline` edge case works; cached vs uncached agree across the range.
- Runic-formatted.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,6 +10,7 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [weakdeps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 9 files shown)


ChrisRackauckas-Claude:
Added a follow-up commit: **Make BSpline evaluation allocation-free as well as thread-safe**.

The original #532 fix made BSpline eval reentrant by allocating a per-call scratch buffer (`zeros(eltype(t), n)`), which removed the data race but added a heap allocation per call (848 B for a length-99 scalar spline). QuadraticSpline was already allocation-free via its local-variable rewrite; this brings BSpline to parity.

**Approach:** a degree-`d` B-spline has only `d + 1` nonzero basis functions at any point, so the new `bspline_nonzero_coefficients` computes them into a stack-allocated `MVector` (StaticArrays) and returns an `SVector` window + control-point offset. Allocation-free *and* reentrant — no shared or heap state per call. All four `_interpolate` and four `_derivative` BSpline methods use it.

To keep a single statically-allocation-free path (AllocCheck sees every branch reachable from the call operator, including extrapolation paths into `_derivative`), the heap fallback is removed and degree is bounded at construction (`d < 16`). Max degree used anywhere in this repo is 3; degree ≥16 global B-splines are numerically degenerate.

`StaticArrays` moves from a test-only dep to a hard dep.

**Allocations/call (length-99 scalar spline):**

| | before | after |
|---|---|---|
| BSpline value | 848 B | **0 B** |
| BSpline derivative | 848 B | **0 B** |
| BSplineApprox | 128 B | **0 B** |
| BSpline array value | 1888 B | 1040 B (output array only) |

**Verification (all run locally):**
- New `bspline_nonzero_coefficients` matches the reference `spline_coefficients!` exactly (max diff `0.0`) over degrees 1–7 incl. both endpoints
- Derivatives match finite differences (~1e-10) and ForwardDiff (~1e-16)
- `AllocCheck.check_allocs`: 0 allocation sites for value and derivative
- `test/qa/alloc_tests.jl` (with new BSpline value+derivative cases): 16/16
- `test/Methods/thread_safety_tests.jl` with `-t4`: 28/28 (concurrent stress path exercised)
- `interpolation_tests.jl`: 17412 pass, 0 fail; `derivative_tests.jl`: 34479 pass, 0 fail

Issue #567, opened 2026-07-31
BSplineInterpolation: :Uniform/:Uniform diverges at the boundaries as n grows; :ArcLen/:Average converges at 2nd order regardless of degree


NikoBiele:
Interpolating a smooth function from uniformly spaced samples, I ran into two
distinct problems with `BSplineInterpolation`. Both are visible in the same
sweep, so I am reporting them together, happy to split into two issues if you
prefer.

Test problem throughout: `f(x) = sin(x)` on `[0, 2π]`, sampled at `n` uniformly
spaced points, error measured as `max |A(x) - f(x)|` over a dense grid of 2001
points.

## 1. `:Uniform, :Uniform` diverges as n increases

| n | degree 3 | degree 5 |
|---|---|---|
| 20 | 4.97e-5 | 6.48e-7 |
| 40 | 2.17e-6 | 2.52e-6 |
| 80 | 1.15e-7 | 6.63e0 |
| 160 | 4.79e-6 | 6.81e18 |
| 320 | 8.49e5 | 2.69e55 |
| 640 | 5.81e27 | `SingularException(639)` |

Adding samples makes the interpolant worse, by many orders of magnitude per
refinement, and eventually the construction fails outright. Onset is
degree-dependent: degree 5 breaks from n=80, degree 3 from n=160.

### The failure is a boundary layer

Plotting the pointwise error rather than its maximum shows the interior is
healthy and the problem is entirely at the ends. At n=640, degree 3, the
interior error sits at a low level across the domain while both endpoints reach
~1e20, decaying exponentially inward over a fixed number of knot spans:

<img width="600" height="400" alt="Image" src="https://github.com/user-attachments/assets/831e7c4c-c7db-4acf-8fd5-c3888971ff27" />

Straight lines on a log axis mean geometric decay per knot span. This is
consistent with the boundary block of the collocation system being nearly
singular: the amplified end-condition error propagates inward through the
decaying root of the B-spline symbol. The interior collocation and the basis
itself appear to be fine: an interior-only error norm converges normally.

The `SingularException` at large n looks like the same conditioning failure once
it becomes fatal rather than merely large.

## 2. `:ArcLen, :Average` is stable but only 2nd order, for any degree

| n | degree 3 | degree 5 | ratio (degree 3) |
|---|---|---|---|
| 20 | 3.19e-3 | 3.30e-3 | — |
| 40 | 7.69e-4 | 7.73e-4 | 4.15 |
[...]

PR #568, opened 2026-07-31
Invert the B-spline parameterization exactly when evaluating

ChrisRackauckas-Claude:
**Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

Fixes the second half of https://github.com/SciML/DataInterpolations.jl/issues/567. The first half (`:Uniform` knot conditioning) is a separate, stacked PR.

## The problem

`BSplineInterpolation` fits `u` against a curve parameter `p ∈ [0, 1]`, not against `t`. Evaluating `A(t)` therefore has to invert the map `p ↦ t`. `interpolation_methods.jl` did that by linear interpolation of the `(t[i], p[i])` pairs:

```julia
t = A.p[idx] + (t - A.t[idx]) / (A.t[idx + 1] - A.t[idx]) * (A.p[idx + 1] - A.p[idx])
```

That inversion is only 2nd order accurate, so it caps the entire interpolant at O(h²) *regardless of the requested degree*, whenever `p` is not affine in `t`. Two configurations are affected:

- `pVecType = :ArcLen` — always, since chord length is a nonlinear function of `t`.
- `pVecType = :Uniform` — whenever `t` is not uniformly spaced.

The reporter's diagnostic is what makes it unambiguous: for `:ArcLen`/`:Average` the errors for degree 3 and degree 5 agreed to three significant figures at every `n`, i.e. the degree was not reaching the result at all.

The spline itself was never wrong. Error measured *at the data points* is 1e-16 at every `n`; all the error is between them, and it converges at exactly order 2:

```
:ArcLen/:Average, degree 5, f = sin

     n       at nodes   at midpoints  ratio
    20      1.110e-16      3.295e-03
   160      2.220e-16      4.648e-05  4.051
   640      3.331e-16      2.878e-06  4.012
```

## The fix

Fit `t` against `p` with the same basis and knots, and invert that spline properly.

- The `t`-coordinate control points `ct` come from the collocation matrix that is already factorized for `c`, so construction costs one extra triangular solve, not another factorization.
- Evaluation solves `T(p) = t` with Newton, seeded by the old linear estimate and bracketed by `[p[idx], p[idx+1]]`. `T` matches the strictly increasing `t` at the bracket ends but need not be monotone in between, so the bracket is maintained and Newton falls back to bisection on any step that would leave it.

Degree `d` now converges at order `d + 1`. Max error of `f(x) = exp(sin 3x)` over `[0, 2π]`, `:ArcLen`/`:Average`:

|   n |  d=3 before |  d=3 after |  d=5 before |  d=5 after |
|----:|------------:|-----------:|------------:|-----------:|
| 160 |    4.24e-03 |   2.14e-03 |    8.50e-03 |   1.34e-02 |
| 320 |    8.32e-04 |   6.92e-05 |    8.87e-04 |   1.25e-04 |
| 640 |    2.07e-04 |   2.09e-06 |    2.07e-04 |   2.50e-07 |

### Differentiability

The final Newton step is taken unguarded, on purpose. Writing the Newton map as `N(p, t) = p - r(p,t)/T′(p)`, we have `∂N/∂p = r·T″/T′²`, which vanishes as the residual does, leaving `∂N/∂t = 1/T′(p)`. So once the value has converged, one more step returns the implicit-function derivative rather than the derivative of the iteration — no unrolling-error decay, and no need to strip `Dual`s (which would not have worked for `Unitful`/`Symbolics` anyway).

Verified: `derivative(A, x)` agrees with `ForwardDiff.derivative(A, x)` to 1.7e-14 on the inverting path.

### Cost

When `p` is affine in `t` — `:Uniform` parameters on uniform `t`, the common case — the closed form is already exact, so it is detected at construction (`affine_param`) and the inversion is skipped entirely.

| path | ns/eval | allocs |
|---|---:|---:|
| affine (unchanged) | 147 | 0 |
| inverting | 611 | 0 |

## ⚠️ Behaviour change — needs a version-bump decision

**`A(t)` returns different values than before for any non-affine parameterization.** This is not limited to `:ArcLen`; it includes `:Uniform` parameters on non-uniformly spaced `t`, which is common. Values *at* the data points are unchanged (still machine-precision exact).

Four hardcoded regression values in `test/interpolation_tests.jl` changed as a result. Because those are lock-in values on arbitrary data with no analytic ground truth, I did not simply paste the new output: I checked each against an **independent implementation** — a from-scratch Cox–de Boor basis, a `BigFloat` linear solve, and plain bisection (no Newton) — which reproduces the new values to 1–2 ulp:

```
--- Uniform / Uniform, degree 2 ---
  t=  25.0  package=13.595900807689079  independent=13.595900807689077  |diff|=1.78e-15
  t=  80.0  package=10.282089579350609  independent=10.282089579350608  |diff|=1.78e-15
  t= 190.0  package=14.193576669619798  independent=14.193576669619800  |diff|=1.78e-15
  t= 225.0  package=11.018696916649047  independent=11.018696916649047  |diff|=0.00e+00
--- ArcLen / Average, degree 2 ---
  t=  25.0  package=13.362987841699956  independent=13.362987841699956  |diff|=0.00e+00
  t=  80.0  package=10.685461155425084  independent=10.685461155425083  |diff|=1.78e-15
  t= 190.0  package=13.436601114574380  independent=13.436601114574382  |diff|=1.78e-15
  t= 225.0  package=11.366765953659648  independent=11.366765953659646  |diff|=1.78e-15
```

The `BSplineInterpolation` struct also gains two fields (`ct`, `affine_param`), so the undocumented positional inner constructor changed arity.

Please decide whether this ships as v9.1 or v10.

## Tests

New `@testset "Parameterization inversion accuracy"` covering both affected configurations:

- absolute error thresholds at `n = 100` that the old 2nd-order code could not meet (it sat at ~1.2e-4 for both degrees);
- observed convergence order `> 3.5` for degree 3 and `> 5.0` for degree 5, which is the property that actually regressed;
- interpolation still exact at the data points;
- `affine_param` is exercised in both states.

## Not covered

`BSplineApprox` evaluates through the same linear estimate and has the same cap. Left out to keep this reviewable; happy to follow up. Separately, while testing I noticed `BSplineApprox` with `:Average` knots does not appear to converge at all as the control point count grows (~6e-2 flat for `h = n/2`, `n = 100…400`, degree 3) — that looks like a third, unrelated bug and I have not investigated it.

## Test status

`Core`, `Methods`, `Misc` all pass. `Extensions` and `QA` each have one **pre-existing failure that reproduces identically on an unmodified `master` checkout** — I verified both against a clean worktree rather than assuming:

- `Extensions`: 8 SparseConnectivityTracer "Global Hessian sparsity" errors (`iszero` on a `HessianTracer` inside `FillArrays._foldl_length_op`). Master baseline is `592 passed, 0 failed, 8 errored`; this branch is byte-identical at `592 passed, 0 failed, 8 errored`. Version-sensitive — it passes under some resolved SCT/FillArrays combinations.
- `QA`: `all_qualified_accesses_are_public` fails because `require_one_based_indexing` is not public in `Base` but is imported at `src/interpolation_utils.jl:308`. Same on master.

Neither is touched by this PR. Happy to open separate issues for them.

https://claude.ai/code/session_01Nh5RgcPFZbKDNh1y61ZEAe

### DIFF
--- a/src/DataInterpolations.jl
+++ b/src/DataInterpolations.jl
@@ -4,7 +4,7 @@ module DataInterpolations
 
 abstract type AbstractInterpolation{T} end
 
-using LinearAlgebra: LinearAlgebra, Tridiagonal, dot, norm, normalize!
+using LinearAlgebra: LinearAlgebra, Tridiagonal, dot, lu, norm, normalize!
 using RecipesBase: RecipesBase, @recipe, @series
 using PrettyTables: PrettyTables, pretty_table
 using ForwardDiff: ForwardDiff
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -254,8 +254,7 @@ function _derivative(A::BSplineInterpolation{<:AbstractVector{<:Number}}, t::Num
     t > A.t[end] && return zero(A.u[end])
     idx = get_idx(A, t, iguess)
     n = length(A.t)
-    scale = (A.p[idx + 1] - A.p[idx]) / (A.t[idx + 1] - A.t[idx])
-    t_ = A.p[idx] + (t - A.t[idx]) * scale
+    t_, scale = bspline_param_and_scale(A, t, idx)
     ducum = zero(eltype(A.u))
     if t == A.t[1]
         ducum = (A.c[2] - A.c[1]) / (A.k[A.d + 2])
@@ -280,8 +279,7 @@ function _derivative(
     t > A.t[end] && return zeros(size(A.u)[1:(end - 1)]...)
     idx = get_idx(A, t, iguess)
     n = length(A.t)
-    scale = (A.p[idx + 1] - A.p[idx]) / (A.t[idx + 1] - A.t[idx])
-    t_ = A.p[idx] + (t - A.t[idx]) * scale
+    t_, scale = bspline_param_and_scale(A, t, idx)
     ducum = zeros(size(A.u)[1:(end - 1)]...)
     if t == A.t[1]
         ducum = (A.c[ax_u..., 2] - A.c[ax_u..., 1]) / (A.k[A.d + 2])
--- a/src/interpolation_caches.jl
+++ b/src/interpolation_caches.jl
@@ -928,6 +928,11 @@ end
 It is a curve defined by the linear combination of `n` basis functions of degree `d` where `n` is the number of data points. For more information, refer [https://pages.mtu.edu/~shene/COURSES/cs3621/NOTES/spline/B-spline/bspline-curve.html](https://pages.mtu.edu/%7Eshene/COURSES/cs3621/NOTES/spline/B-spline/bspline-curve.html).
 Extrapolation is a constant polynomial of the end points on each side.
 
+The curve is parameterized by `p ∈ [0, 1]` rather than by `t`, so both `u` and `t` are
+fitted against `p` and evaluating at a time inverts the `p ↦ t` spline. When `p` is
+affine in `t` — `:Uniform` parameters on uniformly spaced `t` — that inversion is
+closed-form; otherwise it costs a few Newton steps per evaluation.
+
 ## Arguments
 
   - `u`: data points.
@@ -951,7 +956,7 @@ Extrapolation is a constant polynomial of the end points on each side.
 
 """
 struct BSplineInterpolation{
-        uType, tType, pType, kType, cType, scType, T, propsType,
+        uType, tType, pType, kType, cType, ctType, scType, T, propsType,
     } <:
     AbstractInterpolation{T}
     u::uType
@@ -960,6 +965,8 @@ struct BSplineInterpolation{
     p::pType  # params vector
     k::kType  # knot vector
     c::cType  # control points
+    ct::ctType  # control points of the p -> t map, used to invert it
+    affine_param::Bool  # whether p is affine in t, making that inversion trivial
     sc::scType  # Spline coefficients (preallocated memory)
     pVecType::Symbol
     knotVecType::Symbol
@@ -975,6 +982,8 @@ struct BSplineInterpolation{
             p,
             k,
             c,
+            ct,
+            affine_param,
             sc,
             pVecType,
             knotVecType,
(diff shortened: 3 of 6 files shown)


ChrisRackauckas-Claude:
CI triage: the 3 red checks here are pre-existing on `master`, not introduced by this PR.

- `master`'s own latest `Tests.yml` run ([5cd871e8, 2026-07-28](https://github.com/SciML/DataInterpolations.jl/actions/runs/30364050598)) fails on exactly `Extensions (julia 1)` and `Extensions (julia lts)` — the same two jobs.
- `master`'s `Downgrade.yml` has been failing since at least 2026-07-22 ([cac5a512](https://github.com/SciML/DataInterpolations.jl/actions/workflows/Downgrade.yml)).
- All 6 failing jobs across both PRs report the identical `592 passed, 0 failed, 8 errored` — 8 SparseConnectivityTracer "Global Hessian sparsity" errors (`iszero` on a `HessianTracer` inside `FillArrays._foldl_length_op`). I reproduced the same `592/8` on a clean local `master` worktree.

Happy to open a separate issue for the SCT breakage if useful. Two other pre-existing problems I noticed while testing, also unrelated to this PR:

- `all_qualified_accesses_are_public` fails locally because `require_one_based_indexing` is not public in `Base` but is imported at `src/interpolation_utils.jl:308`. (CI's QA job passes, so it appears version-dependent on ExplicitImports.)
- The docs `linkcheck` intermittently fails on `http://www.cad.zju.edu.cn/home/zhx/GM/009/00-bsia.pdf` in the `BSplineApprox` docstring (`src/interpolation_caches.jl:1205`) — that host looks dead.

ChrisRackauckas-Claude:
Superseded by #511, which removed the intermediate \`p ∈ [0,1]\` parameterization entirely rather than inverting it. With no \`t → p\` map there is nothing left to invert, and \`pVecType\`/\`:ArcLen\` are no longer part of the constructor signature.

Verified on master (c37fceb) that the symptom this PR targeted is gone — `f = sin` on `[0, 2π]`, `:Average` knots, max error over 2001 points:

| n | d=3 | ratio | d=5 | ratio |
|---|---|---|---|---|
| 80 | 1.04e-7 | 30.2 | 5.00e-10 | 138.4 |
| 160 | 6.35e-9 | 16.4 | 3.75e-12 | 133.3 |
| 320 | 3.92e-10 | 16.2 | 2.86e-14 | 130.8 |

Degree 3 converges at ratio 16 = O(h^4) and degree 5 reaches machine precision, against the flat ratio 4 for both degrees that this PR was written to fix.

This branch also no longer applies: it adds `ct` and `affine_param` fields to a struct #511 rewrote.

The second half of #567 — the `:Uniform` knot conditioning blowup — is unaffected by #511 and still reproduces on master. That is being reopened separately against the new code.

Issue #572, opened 2026-08-03
ConstantInterpolation: integral disagrees with the interpolation for Linear (NaN) and Extension extrapolation


ChrisRackauckas-Claude:
`integral` of a `ConstantInterpolation` disagrees with the interpolation itself outside the data range, for two of the extrapolation types:

  - `ExtrapolationType.Linear` returns `NaN` on both sides, for both `dir` settings.
  - `ExtrapolationType.Extension` integrates the wrong interval's value on one side: the right side for `dir = :left`, the left side for `dir = :right`.

`ExtrapolationType.Constant`, `Periodic` and `Reflective` are fine. This is independent of the shape of `u` — it reproduces identically for vector and matrix `u`.

## Reproduction

On master (`c37fceb`), Julia 1.11.9:

```julia
using DataInterpolations, QuadGK

u = [1.0, 2.0, 5.0]
t = [1.0, 2.0, 3.0]

for dir in (:left, :right),
    extrapolation in (
        ExtrapolationType.Constant, ExtrapolationType.Linear, ExtrapolationType.Extension,
    )

    A = ConstantInterpolation(u, t; extrapolation, dir)
    for (side, t1, t2) in (("left", -1.0, 1.0), ("right", 3.0, 5.0))
        got = DataInterpolations.integral(A, t1, t2)
        want = quadgk(A, t1, t2; atol = 1.0e-12, rtol = 1.0e-12)[1]
        println(rpad("dir=$dir $extrapolation $side", 40), rpad(got, 6), " vs ", want)
    end
end
```

```
dir=left Constant left                  2.0    vs 2.0
dir=left Constant right                 10.0   vs 10.0
dir=left Linear left                    NaN    vs 2.0      <- wrong
dir=left Linear right                   NaN    vs 10.0     <- wrong
dir=left Extension left                 2.0    vs 2.0
dir=left Extension right                4.0    vs 10.0     <- wrong
dir=right Constant left                 2.0    vs 2.0
dir=right Constant right                10.0   vs 10.0
dir=right Linear left                   NaN    vs 2.0      <- wrong
dir=right Linear right                  NaN    vs 10.0     <- wrong
dir=right Extension left                4.0    vs 2.0      <- wrong
dir=right Extension right               10.0   vs 10.0
```

`quadgk` integrates `A` itself, so the reference column is what the interpolation object actually evaluates to.

## Cause

[...]

PR #563, opened 2026-07-24
Adding matrix u extrapolation methods for basic interpolation schemes

jbiffl:
This PR adds support for extrapolation with matrix u for the following interpolation methods that already support interpolation with matrix u:

- `ConstantInterpolation()` using `ExtrapolationType.Constant`
- `LinearInterpolation()` using `ExtrapolationType.Constant`, `ExtrapolationType.Linear`, and `ExtrapolationType.Extension`
- `QuadraticInterpolation()` using `ExtrapolationType.Constant`, `ExtrapolationType.Linear`, and `ExtrapolationType.Extension`

It also adds supporting derivative and integral methods and tests for the above.

It should fix https://github.com/SciML/DataInterpolations.jl/issues/561

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API

PR #573, opened 2026-08-03
Matrix u extrapolation without the extra allocations (continues #563)

ChrisRackauckas-Claude:
> **Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

Continues https://github.com/SciML/DataInterpolations.jl/pull/563 by @jbiffl (matrix `u` support for `ConstantInterpolation`, `LinearInterpolation` and `QuadraticInterpolation` extrapolation, fixing #561). That PR's commits are included unchanged and rebased onto current master; this adds the follow-up asked for in https://github.com/SciML/DataInterpolations.jl/pull/563#discussion_r3700440903 ("You don't need to slice that?").

## The problem

Reaching a data point out of a matrix `u` was done by copying a column — `u[:, idx]`, `u[:, 1]`, `u[:, end]` — and the arithmetic consuming it was left unfused, so each intermediate materialised its own array. A single extrapolated value cost up to 15 arrays where 1 is needed.

## After

Measured with `@allocated`, `u = [1.0 3.0 2.0; 2.0 5.0 7.0]`, `cache_parameters = true`, Julia 1.11. One result column is 80 bytes.

| | eval | derivative | integral |
| --- | --- | --- | --- |
| #563 as-is | 240–1200 B | 240–880 B | 880–2768 B |
| this PR | 80–160 B | 0–80 B | 240–400 B |

`eval` and `derivative` are down to the single result array, except where a parameter itself has to be computed (`QuadraticInterpolation`'s slope under `Constant`/`Linear` extrapolation) — that is a second genuinely-computed value, not a copy. `integral` composes 3–5 array-valued terms and allocates one array each; going below that would need an in-place `_integral!`, which is a larger change.

Vector `u` is unaffected: still 0 allocations, and the same runtime as master (BenchmarkTools sweep over 997 query points for `Constant`/`Linear`/`Quadratic` × `cache_parameters`, both trees, differences within noise).

## What changed

  - `_u_view` reaches the data point at an index without copying (`selectdim` on the last dimension, or plain indexing when `u` holds one value per point). `_first`/`_last` are defined in terms of it. `_u_point` is the copying variant, used only where the value is returned to the caller — `ConstantInterpolation`'s extrapolation, which must not hand back a view into `u`.
  - The expressions that consume them are fused (`@.`) in `_extrapolate_left`/`_extrapolate_right`, `_extrapolate_derivative_*`, `_extrapolate_integral_*`, `_interpolate` and `_derivative`.
  - The duplicated `_integral` methods are gone. `{<:AbstractVector{<:Number}}` and `{<:AbstractMatrix{<:Number}}` differed only in how they indexed the data point, so each collapses into one `{<:AbstractArray{<:Number}}` method — same type coverage, no slice.
  - `integral` accumulates into its own result array rather than allocating one per interval.
  - `linear_interpolation_parameters` and `quadratic_interpolation_parameters` take views instead of copies.
  - `SmoothedConstantInterpolation`'s linear right-extrapolation branch computed a value and discarded it before delegating; removed.

## Fixes along the way

  - `integral(A, t, t)` on array-valued `u` called `zero(eltype(A.I))`, where `eltype(A.I)` is an array type — `MethodError`. It now returns a zero array of one data point's shape.
  - Left/right extrapolation for `u` with more than two dimensions did `first(A.u) + slope * Δt`, adding a scalar to an array. This errored on master too; `LinearInterpolation(rand(2, 3, 4), t; extrapolation = ExtrapolationType.Linear)(0.0)` now works, as does `integral` on it.

## Tests

  - `integral(A, t, t)` for matrix `u` in the `ConstantInterpolation` and `LinearInterpolation` sets.
  - An aliasing check that `ConstantInterpolation`'s extrapolated value does not share memory with `u`.
  - `"Matrix u extrapolation allocations"`: budgets each of eval/derivative/integral against the byte cost of one result array, so a reintroduced slice or unfused broadcast fails the suite. Budgets verified on 1.10, 1.11 and 1.12 — the byte counts are identical across all three.
  - Removed the commented-out `test_extrapolation(ConstantInterpolation, u, t)` and the stray assignments/trailing whitespace from #563.

`test_extrapolation(ConstantInterpolation, u, t)` still cannot be enabled, but not because of anything here: `ConstantInterpolation`'s `integral` disagrees with the interpolation outside the data range on master, for vector `u` as much as matrix `u`. Filed separately as https://github.com/SciML/DataInterpolations.jl/issues/572.

## Checklist

  - [x] The new code follows the [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC)

### DIFF
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -57,9 +57,9 @@ function _extrapolate_derivative_left(A, t, order)
     return if extrapolation_left == ExtrapolationType.None
         throw(LeftExtrapolationError())
     elseif extrapolation_left == ExtrapolationType.Constant
-        zero(first(A.u) / one(A.t[1]))
+        zero(_first(A.u) / one(A.t[1]))
     elseif extrapolation_left == ExtrapolationType.Linear
-        (order == 1) ? _derivative(A, first(A.t), 1) : zero(first(A.u) / one(A.t[1]))
+        (order == 1) ? _derivative(A, first(A.t), 1) : zero(_first(A.u) / one(A.t[1]))
     elseif extrapolation_left == ExtrapolationType.Extension
         (order == 1) ? _derivative(A, t, length(A.t)) :
             ForwardDiff.derivative(
@@ -93,10 +93,10 @@ function _extrapolate_derivative_right(A, t, order)
     return if extrapolation_right == ExtrapolationType.None
         throw(RightExtrapolationError())
     elseif extrapolation_right == ExtrapolationType.Constant
-        zero(first(A.u) / one(A.t[1]))
+        zero(_first(A.u) / one(A.t[1]))
     elseif extrapolation_right == ExtrapolationType.Linear
         (order == 1) ? _derivative(A, last(A.t), length(A.t)) :
-            zero(first(A.u) / one(A.t[1]))
+            zero(_first(A.u) / one(A.t[1]))
     elseif extrapolation_right == ExtrapolationType.Extension
         (order == 1) ? _derivative(A, t, length(A.t)) :
             ForwardDiff.derivative(
@@ -251,7 +251,7 @@ function _derivative(A::AkimaInterpolation{<:AbstractVector}, t::Number, iguess)
 end
 
 function _derivative(A::ConstantInterpolation, t::Number, iguess)
-    return zero(first(A.u))
+    return zero(_first(A.u))
 end
 
 function _derivative(A::ConstantInterpolation{<:AbstractVector}, t::Number, iguess)
--- a/src/derivatives.jl
+++ b/src/derivatives.jl
@@ -52,14 +52,21 @@ function derivative(A, t, order = 1)
     end
 end
 
+# Zero of the derivative's value type, shaped like one data point.
+function _zero_derivative(A)
+    u₁ = _first(A.u)
+    t₁ = one(A.t[1])
+    return @. zero(u₁ / t₁)
+end
+
 function _extrapolate_derivative_left(A, t, order)
     (; extrapolation_left) = A
     return if extrapolation_left == ExtrapolationType.None
         throw(LeftExtrapolationError())
     elseif extrapolation_left == ExtrapolationType.Constant
-        zero(_first(A.u) / one(A.t[1]))
+        _zero_derivative(A)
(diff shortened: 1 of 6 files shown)


ChrisRackauckas (approved):

ChrisRackauckas-Claude:
The Julia 1.10 failures on the first run (`Core (julia lts)` and `Downgrade`, which also runs 1.10) were inherited from #563, not introduced here. I reproduced them on an unmodified checkout of #563 at c37fceb-era master: same three errors, same lines, same inferred type.

Cause: giving `_integral` methods for array-valued `u` means `cumulative_integral`'s element type now genuinely depends on inferring them, where before there was no matching method and it inferred trivially as `Vector{Union{}}`. Julia 1.10 doesn't get all the way through that chain; 1.11 and later do.

I tried three formulations before concluding it isn't fixable at the source level:

  - `_u_view` via `view(u, :, idx)` / `ntuple(Val(N-1))` instead of `selectdim` — no change.
  - `linear_interpolation_parameters` restored to master's single-method shape with views instead of the branch-of-broadcasts — no change.
  - `cumulative_integral` built by broadcast rather than `cumsum` over a generator, to avoid `@default_eltype`'s nested `return_type` call — strictly worse (4 errors, and the whole `I` type widened rather than just its element type).

So the last commit moves the three 1.10 `broken` markers to where they now apply rather than papering over anything. It is not a net loosening on 1.10: `QuadraticInterpolation` with matrix `u` now infers where it previously could not, so that marker is removed and the assertion is now real; `LinearInterpolation` with matrix `u` swaps which of the two knot types trips it, so that one marker moves from the `AbstractRange` case to the `Vector` case. One marker gone, one moved, none added. Julia 1.11+ is unaffected — both conditions were already `VERSION < v\"1.11\"`-gated.

Verified locally: `GROUP=Core` passes on 1.10, 1.11 and 1.12; the other groups pass on 1.11 and QA on 1.12.

`Downgrade` will still be red — it is red on master too, with the pre-existing `NonlinearCurveFitProblem` `MethodError` at `interpolation_tests.jl:1616` (unrelated to this PR).

PR #577, opened 2026-08-04
refactor: add missing features

sathvikbhagavan:
This PR does:

- Adds full Matrix/Vector{Vector} support (construct, eval, derivative, integral) to Akima, QuadraticSpline, CubicSpline, CubicHermiteSpline, QuinticHermiteSpline, SmoothedConstantInterpolation, BSplineInterpolation/BSplineApprox, and closes remaining integral/derivative gaps in Linear/Quadratic/ConstantInterpolation.
- Adds an analytical integral and Vector{Vector} construction support to SmoothArcLengthInterpolation, and fixes its cache_parameters kwarg, which was silently ignored (always built as if false).
- Fixes a few bugs: broken integral caching for Vector{Vector} data, an integrate_cubic_polynomial array-arithmetic bug, wrong extrapolation indexing in SmoothedConstantInterpolation, a single-point ConstantInterpolation edge case, and a SmoothArcLengthInterpolation aliasing bug where A.(t_vec) (broadcasting) and the documented Vector{Vector}-out in-place form silently returned the same last-evaluated buffer for every element instead of distinct values (root cause: an in_place buffer-reuse kwarg unique to this type, now removed so it behaves like every other interpolation).
- Fixes the Plots.jl recipe, which silently produced garbage (thousands of degenerate one-point series) or errored outright for any Matrix/Vector{Vector}-valued interpolation — plotting effectively never worked for multi-dimensional data before this. Also fixes a related regression risk for BSplineApprox specifically: its "Data points" series must show the true raw data, not the smoothed curve re-evaluated at the knots.
- Fixes show/pretty-printing, which crashed for higher-rank (AbstractArray{T,3}) data and silently mis-rendered Vector{Matrix} data.
- Rewrites SmoothArcLengthInterpolation's _interpolate/_derivative to avoid in-place broadcast assignment, fixing Zygote/Mooncake/ForwardDiff differentiability w.r.t. t (previously failed unconditionally with "Mutating arrays is not supported").
- Adds Zygote/Mooncake AD test coverage for all the new shape-support code paths, plus new t-derivative coverage for SmoothArcLengthInterpolation (its u-gradient through the constructor remains unsupported by both frameworks — documented, not fixed, since it requires rewriting the mutation-heavy geometry-fitting algorithm).
- Documents all of the above in the README (new Feature Support and Automatic Differentiation tables) and docs (SmoothArcLengthInterpolation added to the interpolation listing, tutorial simplified now that Vector{Vector} input no longer needs manual hcat).
- Full test suite passes with zero failures/errors; verified consistent results across shapes, cache_parameters, and all extrapolation modes.

Made with Claude Code

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,32 @@ corresponding to `(u,t)` pairs.
   - `CubicHermiteSpline(du, u, t)` - A third order Hermite interpolation, which matches the values and first (`du`) order derivatives in the data points exactly.
   - `PCHIPInterpolation(u, t)` - a type of `CubicHermiteSpline` where the derivative values `du` are derived from the input data in such a way that the interpolation never overshoots the data.
   - `QuinticHermiteSpline(ddu, du, u, t)` - A fifth order Hermite interpolation, which matches the values and first (`du`) and second (`ddu`) order derivatives in the data points exactly.
+  - `SmoothArcLengthInterpolation(u)` - A C¹ smooth, unit-speed (arc-length parameterized) interpolation through `u` given in matrix form `(ndim, ndata)`, approximated with line and circle segments. Unlike the methods above, `t` is derived from `u` (the cumulative arc length) rather than supplied by the user.
+
+## Feature Support
+
+Every interpolation method above accepts `u` as a plain `Vector`, a `Matrix` (`ndim × ndata`), or a `Vector` of equal-length `Vector`s — and `derivative`/`integral` work for all three, with two exceptions:
+
+| | `Vector` | `Matrix` | `Vector{Vector}` |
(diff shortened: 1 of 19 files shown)


ChrisRackauckas:
wow, that's a lot 😅  😅  😅  😅  😅  😅  😅  😅  😅  😅  😅  😅  😅

sathvikbhagavan:
Yeah 😅 
I started off with just to add support for higher order arrays for the remaining methods and then went down the rabbit hole. Also, a big chunk of changes are in tests.

ChrisRackauckas:
Jesus, close a bunch of issues?

Issue #580, opened 2026-08-10
Low level Windows crashes since v9.0.0


SouthEndMusic:
I'm experiencing crashes like this without a Julia stacktrace:

```
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7ffd5c3c415a -- RtlVirtualUnwind2 at C:\WINDOWS\SYSTEM32\ntdll.dll (unknown line)
in expression starting at <top level call>
Allocations: 497444091 (Pool: 497426698; Big: 17393); GC: 167
```

I have isolated the introduction of the problem to the `DataInterpolations.jl` update `8.10.0 => 9.0.0`. It's very elusive; I've only been able to trigger the crash in certain `OrdinaryDiffEq.jl` solves which make heavy use of `DataInterpolations.jl`. The point of the crash in the simulation does seem deterministic. Running the solve under `@profview` makes the crash much more likely. It seems that it is caused by a race condition or memory state that only  occurs under very specific circumstances. 

I haven't been able to create a MRE, because changes I make to the run like print statements stop the crash from happening.

SouthEndMusic:
@ChrisRackauckas I have identified https://github.com/SciML/DataInterpolations.jl/pull/531 as the culprit

ChrisRackauckas:
Which search method?

SouthEndMusic:
Could be one of `KIND_LINEAR_SCAN, KIND_BRACKET_GALLOP, KIND_UNIFORM_STEP`

ChrisRackauckas:
Can you force the kind and see which one it reproduces with?

SouthEndMusic:
Yep, it reproduces with `KIND_LINEAR_SCAN`, not with `KIND_BRACKET_GALLOP`. I didn't test `KIND_UNIFORM_STEP`, as not all interpolations are uniformly spaced.

ChrisRackauckas-Claude:
[...]

PR #582, opened 2026-08-12
Avoid KIND_LINEAR_SCAN on short non-uniform interpolations (#580)

ChrisRackauckas-Claude:
Please ignore this PR until it is reviewed by @ChrisRackauckas.

## What changed and why

`FindFirstFunctions.Auto` picks `KIND_LINEAR_SCAN` for non-uniform knots with `length(t) ≤ 16`. That path is the one isolated as the Windows `EXCEPTION_ACCESS_VIOLATION` in `ntdll!RtlVirtualUnwind2` when interpolations are evaluated from an OrdinaryDiffEq Dual RHS ([#580](https://github.com/SciML/DataInterpolations.jl/issues/580)). `KIND_BRACKET_GALLOP` does not reproduce.

`_resolve_strategy_kind` now remaps `KIND_LINEAR_SCAN` → `KIND_BRACKET_GALLOP`. That is the v8 search default for these interpolations. Uniform knots still use `KIND_UNIFORM_STEP`.

This is a workaround, not a root-cause fix of the Windows unwind crash. The linear-scan kernel itself matches `Base.searchsortedlast` (500-case fuzz, 0 mismatches) and is in-bounds under `--check-bounds=yes`. The crash looks like a codegen / unwind-info Heisenbug (`@profview` makes it more likely; prints hide it). We do not have a Windows box here, so we ship the path SouthEndMusic already confirmed is safe.

Standalone MWE (Ribasim-shaped: 40 basins, 8-knot profiles, `invert_integral`, PCHIP Q-H, Rosenbrock + ForwardDiff): https://github.com/SciML/DataInterpolations.jl/issues/580#issuecomment-5273108807

## Verification

Discriminating test (short non-uniform `Linear` / `Constant` / `PCHIP` / `invert_integral` must be `KIND_BRACKET_GALLOP`):

- **Before** (master, no remap): `KIND_LINEAR_SCAN === KIND_BRACKET_GALLOP` fails 4 times.
- **After**: 8/8 pass; kinds are `KIND_BRACKET_GALLOP`. Range knots still resolve to `KIND_UNIFORM_STEP`.

Local (Julia 1.12.6):

```
GROUP=Core  Pkg.test()  → interpolation_tests 17436 pass, 5 pre-existing broken; Core all green
GROUP=QA    Pkg.test()  → alloc_tests 16 pass; qa.jl 20 pass
runic -c on the touched .jl files: clean
typos on the diff: clean
```

## What I did not verify

- The Windows AV itself. This machine is Linux. I cannot show the crash disappearing.
- `GROUP=Methods`, `GROUP=Extensions`, GPU, downstream (Ribasim). Methods was started locally; say so if it is still running when this is filed.
- Whether `@noinline` on the FFF linear-scan kernel would also stop the AV while keeping the algorithm. That is a plausible follow-up in FindFirstFunctions once someone can test on Windows.

## Judgment call

Remapping rather than trying to keep LINEAR_SCAN via `@noinline`. Evidence for the remap: SouthEndMusic reproduced with LINEAR_SCAN and not with BRACKET_GALLOP. Evidence for `@noinline` as a complete fix: none on this machine.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
   - Every interpolation constructor accepts a `search_properties::Union{Nothing, FindFirstFunctions.SearchProperties}` keyword. The default `nothing` probes `t` at construction; passing a pre-built `SearchProperties` skips the probe (useful when constructing many interpolations over the same knot vector).
 
-  - Knot search is dispatched through `FindFirstFunctions.Auto(t)` resolved at construction: uniformly-spaced knots (any `AbstractRange`, or vectors detected as exactly uniform) use a closed-form O(1) lookup; short non-uniform knot vectors use a linear scan; everything else keeps the previous bracketed gallop.
+  - Knot search is dispatched through `FindFirstFunctions.Auto(t)` resolved at construction: uniformly-spaced knots (any `AbstractRange`, or vectors detected as exactly uniform) use a closed-form O(1) lookup; non-uniform knot vectors use the previous bracketed gallop. `KIND_LINEAR_SCAN` (Auto's pick for `length(t) ≤ 16`) is remapped to `KIND_BRACKET_GALLOP` because the linear-scan kernel crashes on Windows (`RtlVirtualUnwind2`) when evaluated from an OrdinaryDiffEq Dual RHS ([#580](https://github.com/SciML/DataInterpolations.jl/issues/580)).
 
   - `LinearInterpolation` with uniformly-spaced knots and floating-point values takes a statically-dispatched fast path (closed-form index + lerp, verified against the live knots) — 5-10x faster per query on uniform grids.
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
`GROUP=Methods` finished locally (Julia 1.12.6):

- `derivative_tests.jl`: 34479 pass
- `integral_inverse_tests.jl`: 317 pass
- `integral_tests.jl`: 8227 pass
- `online_tests.jl`: 30 pass
- `thread_safety_tests.jl`: 21 pass (concurrent stress skipped; this process was single-threaded)

`GROUP=Extensions` and docs were not run. GPU/downstream not run.

ChrisRackauckas-Claude:
Closing: remapping away from `KIND_LINEAR_SCAN` was the wrong fix. The linear-scan kernel itself needs to be corrected (the `@inbounds if` + `&& return` walk is the Windows `RtlVirtualUnwind2` suspect). Follow-up will be in FindFirstFunctions.

PR #588, opened 2026-08-24
refactor: support for VoV and matrix for PCHIPInterpolation

sathvikbhagavan:
Follow up on #577


sathvikbhagavan:
@ChrisRackauckas, after this PR, we can have a major release?

